PR #80, opened 2021-11-02
Small generic strategic case study

Andresj89:
## Summary
This PR adds a small strategic case study that converges to optimality with `CBC` within a few seconds. The case was built based on `input_data_generic_strategic_case_study_LAYFLAT_FULL.xlsx`. The set for pipeline diameters was reduced to one element `D0`, and the optimal solution of the full case study was used to populate the parameter for the initial pipeline capacity.


drouvenm (commented):
Can you replace me as a reviewer by @melody-shellman or @NienkeWagenaar? I think they should be the ultimate "judges" of whether this works as expected or not ;-)

melody-shellman (approved):
This is great! Ran it on my end and it solved in the 2min time frame allotted. Woohoo!

I have a note/question. When the model runs, the last part of the output says the number of solutions is 0. (see screenshot). This comes from the results.write() [ line 106] of the run_strategic_model.py. This same line also prints "Model was solved to optimality". These seem to contradict each other.  Do you know why it say there are 0 solutions if it also says the model was solved to optimality?

image

Andresj89:
@melody-shellman Thanks for checking this. Yes, the model solved to optimality the results are stored in the model instance but are not passed as an output to `results`, accordingly to what I read, this is expensive to do. There is a way to load the solution to results:
```
results = opt.solve(model)
model.solutions.store_to(results)
print(results)
```
But it's not really necessary as long as we can access the solution through the model itself.

NienkeWagenaar (commented):
I think the file looks really good! 
2 questions / comments on my end:
- Does this input file include all the data changes based on Karen's feedback? I see some differences in e.g. the TruckingHourlyCost tab.
- I would propose to include 1 additional pipeline option. Since it it the strategic model, I think this is one of the key decisions.

Andresj89:
This will be addressed in #81

PR #81, opened 2021-11-03
Updates for the strategic model

Andresj89:
## Summary
This PR updates the strategic model as following:

**Input file**
- The tab “PipelineCapacityIncrements” no longer exists
- The tab “PipelineDiameterValues” is added and contains the diameter of the pipeline options in inches
- The tab “Hydraulics” has been added and contains 2 scalars: the roughness and max head loss. 
- The tab “PipelineExpansionCost” has been updated to contain a scalar, representing the cost in $/inchmile.
- The tab “PipelineExpansionDistance” has been added and contains the distance between segments in miles.
- The pipeline distances between locations have been set to 10 miles, this will be reviewed in future updates.

**Run_strategic_model.py**
- Parameter_list is updated to exclude `PipelineCapacityIncrements`, but include `PipelineDiameterValues` and `Hydraulics`
- The constraint `PipelineExpansionCapEx` has been updated accordingly

**Strategic_produced_water_optimization.py**
- Assuming that we might need to precalculate additional parameters later on, the function `_preprocess_data()` was added which takes in `df_Parameters`, and returns it with udated information
- In `_preprocess_data()`, `df_parameters[‘PipelineCapacityIncrements’]` is calculated using Hazen-Williams approximation.
- The storage deposit constraint `StorageDepositCostrule` has been updated. This now includes the trucks from production pads to storage
- Reuse destination delivery constraint `ReuseDestinationDeliveriesRule` has been updated. This constraint now applies only for source locations that are not in the set of fresh water locations:

**Results.py**
- An if-statement was added to the calculation of KPI's to prevent it from throwing an error when a value is of type `None`. In that case it returns a zero.


pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -997,12 +1012,9 @@ def create_model(df_sets, df_parameters, default={}):
...
     model.p_kappa_Pipeline = Param(
-        model.s_L,
-        model.s_L,
-        model.s_D,
-        default=5,
-        initialize=df_parameters["PipelineExpansionCost"],
-        doc="Pipeline construction/expansion capital cost for selected increment [$/bbl]",
+        default=120000,
+        initialize=df_parameters["PipelineExpansionCost"]["pipeline_expansion_cost"],
+        doc="Pipeline construction/expansion capital cost for selected increment [$/inchmile]",
```
drouvenm, line 1017: Can we change this unit to "[$/inch-mile]"? That's how I've seen folks express this.

pareto/tests/test_strategic_model.py
```diff
@@ -120,7 +122,7 @@ def build_strategic_model():
 def test_basic_build(build_strategic_model):
     """Make a model and make sure it doesn't throw exception"""
     m = build_strategic_model
-    assert degrees_of_freedom(m) == 66856
+    assert degrees_of_freedom(m) == 63944
```
drouvenm, line 125: Very nice to see that we are making use of tests now :-)

pareto/utilities/results.py
```diff
@@ -158,31 +158,43 @@ def generate_report(model, is_print=[], fname=None):
```
drouvenm, line 158: General question regarding the results: I've noticed that - for the generic case - most of the produced water now ends up going to disposal (over 88%)... Why is that? Did building pipelines become more expensive? Just seems a bit high and certainly does not align with current industry best practices. Might be worth making some revisions to the data to promote higher reuse rates.

melody-shellman:
## Summary of Changes for Issue #85 
Changes:
•	Add RSA
o	add RSA_Table = {}
o	add model.p_RSA parameter
•	StorageSitePRocessingCapacityRule
o	 Add RSA summation
•	StorageSiteBalanceRule
o	 add RSA summation in BOTH IF and ELSE
•	PipelineCapacityExpansionRule
o	Update SRA to be NOT reversible by default
o	Add constraint for RSA
•	TreatmentBalance
o	 add RSA summation
•	StorageDepostiCostRule
o	add RSA summation
## Testing
•	Run 1: small_strategic_case_study, No code Changes [base case]
•	Run 2: small_strategic_case_study, Add RCA connecions to code 
o	Run 1 and Run 2 had the same results. This is expected, as the input file does not have any valid RCA arcs. 
•	Run 3:  Adjusted the Input file to have valid SCA and RSA arcs.
o	Run 1, 2, 3, have the same results. This makes sense: We wouldn't choose to pipe treated water through storage if we can always pipe directly to completions site. 
•	Run 4: Remove the RCA arcs so treated water MUST go through storage. 
o	The storage sites are used, RSA and SCA pipelines are used!
o	Less water is reused (85% vs 100%)
o	Slack variable for pipeline capacity between storage and completions is triggered [no option to expand pipeline in this small case study]
o	I believe these results make sense

melody-shellman:
## Small_Strategic_Case_Study Notes

- Now includes updated costs from the most recent layflat case (e.g. disposal operational cost, trucking hourly cost)
- With the new cost updates and calculations the model is now slightly slower, allow for at least 5min timeout setting if using cbc
- I tested adding an additional pipeline option as @NienkeWagenaar  suggested, but it did not solve in 10min, so I left it out.  

## Testing Notes
image

pareto/utilities/results.py
Andresj89: Markus, we are going to address this in a future PR, for which we will include the calculation of levelized capital costs for pipelines, this will decrease the cost of a pipeline and promote building pipeline infrastructure to increase water reuse

drouvenm (approved):

MAZamarripa (approved):
lgtm

Issue #88, opened 2021-11-19
Need for Volume- AND Diameter-based Pipeline Data Inputs in Strategic Model


drouvenm:
As discussed, we want/need to have the ability to choose whether existing pipeline capacity is specified in terms of volume (i.e., bbl/day) or in terms of pipeline diameter, segment length, etc.

PR #89, opened 2021-11-19
Pipeline Config Options

melody-shellman:
## Notes
This PR Addresses Issue #88 . 
There are four possible configuration scenarios:

1. User inputs pipeline capacity, pipeline cost is based on capacity
2. User inputs pipeline capacity, pipeline cost is based on distance
3. pipeline capacity is calculated, pipeline cost is based on capacity
4. pipeline capacity is calculated, pipeline cost is based on distance

These four scenarios are handled with two config variables: "pipeline_capacity" and "pipeline_cost"

Note that this PR does not include error handling. For now, a user must include input tabs for all 4 scenarios. Error handling will be in a separate issue. 

## Testing
image

### Addressing Andres' comment: 
- The pipelines between Treatment Facilities and completions pads are the only allowed arcs for recycling.
- The initial capacity of these arcs are 0. These arcs are also the longest. 
- The cost to expand these RCA pipelines for the (distance based, calculated capacity) option is ~350 times larger than the cost to expand the same pipelines for the (capacity based, calculated capacity) option.  
- Because capacity expansion is so expensive, they are not built in the (distance based, calculated capacity) and all produced water must be disposed. 
- The annualization rate PR helped lower the costs, but was still insufficient. I reduced the distances between location (PipelineExpansionDistance). They are now 1/25th of what they used to be and range from 4 to 37 miles instead of 100 to 900 like before. 
- With this update, >80% of demand is met by reused water for all 4 scenarios. 

### Other updates:
- Main is merged in
- Documentation is not yet updated. I plan to update rst file documentation when rst files are added to repository
- Per Miguel's recommendation, testing has been added for all combinations of capex and capacity configurations
- Per Andre's recommendation, model.p_mu_Pipeline is now checked with a CONFIG option 

## Notes on updated 3/2/2022
-Change the PipelineExpansionDistances to larger values (from Andres). The results with larger values are more realistic and provide more flexibility for multi-objective optimization and other developments


NienkeWagenaar (changes requested):
I think we shouldn't upload the results into GIT, as it may cause confusion. I'll add it as discussion item for our coordination meeting.

NienkeWagenaar:
I don't get results for the settings:
Pipeline cost = capacity based
Pipeline capacity = input

When using the layflat input file and running for 1200 seconds. Could someone with a Gurobi solver try this out?

drouvenm:
@NienkeWagenaar I ran this branch with the layflat input file, the pipeline_cost set to capacity_based and the pipeline_capacity set to input. It solved fine for me and Gurobi got to a 6% gap after 60 seconds. Is that helpful to you?

NienkeWagenaar:
> 

@drouvenm, yes that's very helpful, thank you! Do the results make sense? I could have a look at the Excel file, if you like.

Andresj89:
From my tests I think the calculation of capex using `capacity_based` look good. However, for `distance_based` I'm getting strange results. I get the following:
```
Disposal Fraction Produced Water = [%] = 100.00000000000003
Fresh Fraction Completions Demand = [%] = 99.99999999999974
```
I reduced the pipeline Capex in the objective function: `model.v_C_PipelineCapEx*0.1` and `model.v_C_PipelineCapEx*0.01`, but the disposal fraction and fresh water demand was still 100%. I checked the calculation of capacity based on diameters and it is correct.

Also, I think parameter `model.p_mu_Pipeline` should be checked with a CONFIG option too. If the user wants to input capacities and usd/bbl only, then there is no need to provide diameters.

NienkeWagenaar:
> > 
> 
> @drouvenm, yes that's very helpful, thank you! Do the results make sense? I could have a look at the Excel file, if you like.

I confirm the results in the Excel file makes sense. We'll need to discuss the follow up related to the distance_based solutions.

NienkeWagenaar (approved):

pareto/tests/test_strategic_model.py
```diff
@@ -141,6 +145,6 @@ def test_run_operational_model(build_strategic_model):
     # assert results.solver.termination_condition == \
     #     pyo.TerminationCondition.optimal
     # assert results.solver.status == pyo.SolverStatus.ok
-    assert degrees_of_freedom(m) == 66856
+    assert degrees_of_freedom(m) == 63944
     # solutions obtained from running the generic case study
-    assert pytest.approx(1886986000.0, abs=1e-3) == pyo.value(m.v_Z)
+    # assert pytest.approx(380699650.6, abs=1e-1) == pyo.value(m.v_Z)
```
MAZamarripa, line 150: this should not be a comment

MAZamarripa: We need to test all the options

MAZamarripa (changes requested):
looks good
minor comments/changes

Andresj89 (approved):
This looks good.

NienkeWagenaar (approved):
Looks great!

MAZamarripa (approved):

MAZamarripa (approved):
lgtm

PR #92, opened 2021-11-23
Set requirement for Pyomo 6.2 or higher

lbianchi-lbl:


lbianchi-lbl:
Currently `pareto/case_studies/toy_case_study.py` is the only thing preventing being compatible with Pyomo 6.2. @MAZamarripa will look into the best way to deal with this.

Andresj89 (approved):
LGTM

lbianchi-lbl (commented):
Looks good!

PR #96, opened 2021-12-15
Visualization methods

Andresj89:
## Summary

The changes in this PR are:

- Standardized the API for plot_bars and plot_sankey. When reading reports with get_data, the argument "labels" should be included as part of "input_data".
- Added support for data format, we can use "mm-dd-yyyy" or "mm/dd/yyyy". This solves issues when creating plot_bars animations
- Added support to output the plot in any of the following formats: "jpg", "jpeg", "pdf", "png", "svg", "html". Just add the following as part of the additional args: "output_file": "myPlot.html". We can also include a path "output_file": "..\..\myPlot.pdf"
- plot_sankey was revised to speed up the data preprocessing step.


MAZamarripa:
WIP - there is a warning in the plot test.

MAZamarripa (approved):
lgtm

melody-shellman (approved):
These visualizations are cool! I checked that the tests work as expected. Looks good!

PR #98, opened 2022-01-07
stagedfright: Make check for hardcoded data more specific by checking per-file count

lbianchi-lbl:
Following @melody-shellman's feedback after testing stagedfright, she pointed out a limitation in the current version of stagedfright that causes frequent false positives when working with Python files:

- Currently, the check for hardcoded values only takes into account the number of definitions found in the *current* version of the file, without taking into account the hardcoded values that are already present in the file, and that the count is different from file to file
- A more specific test would look at the changes between a reference version of the file (e.g. the one on the `main` branch) and the staged version, and fail the check if the two version differ significantly
  - This functionality would be useful in stagedfright (not only for this specific check, but more in general), but it requires a bit of work to be implemented
- In the meantime, we figured that a usable alternative would be to make the check more specific by keeping a list of the expected number of hardcoded data definition for commonly used files, so that the check only fails if more than the expected number are found (with a configurable tolerance, currently set to 2)
  - This reduces the number of false positives (since the check will only require an allowfile if n > 2 hardcoded data definitions are detected)
  - A possible cause of false negatives is if in the staged version of the file an equal number of hardcoded data definition is removed and added, since at the moment only the total count is considered
    - Possible ways to address this are being considered

Feel free to test this version and see if it addresses the limitations of the current version. In the meantime, I'll keep thinking of more ways to make this check, and other checks on Python files, more specific

### DIFF
--- a/.stagedfright/checks.py
+++ b/.stagedfright/checks.py
@@ -23,8 +23,16 @@ MAP_PATH_EXPECTED_HARDCODED_DATA_COUNT = {
     "pareto/tests/test_operational_model.py": 11,
     "pareto/utilities/get_data.py": 24,
     "pareto/utilities/results": 91,
-    ".stagedfright/checks.py": 7,
 }
+
+# this is meta, two levels deep:
+# - this file is also included checked by stagedfright,
+#   so the hardcoded data used to define this mapping must be counted,
+# - the extra number accounts for constants defined below,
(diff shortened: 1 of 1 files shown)


NienkeWagenaar (approved):

melody-shellman (approved):
Great fix!

NienkeWagenaar (approved):

PR #115, opened 2022-03-28
Node Capacity - New

NienkeWagenaar:
## Summary:

The goal of the changes in this PR is to incorporate node capacities into the strategic model. 
Water that is piped through nodes could be subject to a capacity. In case the user does not enter data, the capacity will be set to infinity.

- ### input_data_generic_strategic_case_study_LAYFLAT_FULL.xlsx:
	- Added 1 tab: NodeCapacities
	- Filled out new tab with dummy data of 1,000,000 bbl/week
- ### small_strategic_case_study.xlsx:
	- Added 1 tab: NodeCapacities
	- Filled out new tab with dummy data of 1,000,000 bbl/week
- ### strategic_produced_water_optimization.py
	- add parameter p_sigma_NetworkNode which is initialized with the node capacity. Default value is set to inf.
	- add node capacity constraint (NetworkNodeCapacityRule)
	- import simple_constraint_rule 
- ### strategic_produced_water_optimization.py
	- add tab NodeCapacities to parameter list
- ### .stagedfright/checks.py
 	- add number of hardcoded values in strategic_produced_water_optimization.py and run_strategic_model.py
- ### tests/test_strategic_model.py
	- add tab NodeCapacities to parameter list

### Legal Acknowledgement

contribution:

   employer.


lbianchi-lbl:
Apparently there's been a breaking change in Click that's causing Black to fail: https://github.com/psf/black/issues/2964. As I understand, a new version of Black has been released that fixes this, so pinning `black==22.3.0` in `requirements-dev.txt` should (hopefully) be enough.

docs/model_library/strategic_water_management/index.rst
```diff
@@ -302,6 +302,8 @@ Strategic Model Mathematical Notation
 
 :math:`\textcolor{green}{σ_{s}^{Processing,Storage}}` =                Weekly processing (e.g. clarification) capacity at storage site
 
+:math:`\textcolor{green}{σ_{n}^{Node}}` =                              Daily capacity per network node
```
melody-shellman, line 305: Because this version of the model is still primarily in bbl/week, I think this should be weekly capacity instead of "daily". This is a good note for me, though-  when I update the unit conversion PR  this should be changed back to daily capacity.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1001,6 +1002,12 @@ def create_model(df_sets, df_parameters, default={}):
         doc="Weekly processing (e.g. clarification) capacity per storage site [bbl/week]",
         mutable=True,
     )
+    model.p_sigma_NetworkNode = Param(
+        model.s_N,
+        default=float("inf"),
+        initialize=df_parameters["NodeCapacities"],
+        doc="Daily capacity per network node [bbl/week]",
```
melody-shellman, line 1009: Similar to above, I think this should be "weekly capacity  per network node [bbl/week]". I will update it "daily" capacity in the unit conversion PR

melody-shellman (changes requested):
I tested a 30,000 bbl/week node capacity for N01 in the small case study, and it works well! Only small change that may be necessary is in the documentation. The parameter is currently "daily" capacity instead of what it actually represents: "weekly" capacity. (Do correct me if I interpreted this incorrectly.)

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1001,6 +1002,12 @@ def create_model(df_sets, df_parameters, default={}):
         doc="Weekly processing (e.g. clarification) capacity per storage site [bbl/week]",
         mutable=True,
     )
+    model.p_sigma_NetworkNode = Param(
+        model.s_N,
+        default=float("inf"),
```
Andresj89, line 1007: I suggest not using `float("inf")` for nodes that don't have a capacity constraint. The reason for this is that the model will be cluttered with equations that are redundant. For example, if node 1 doesn't have a capacity constraint, (and assuming our typical 52 weeks for the planning horizon), then the compiled model will have extra 52 constraints. My suggestion is to create a parameter, equivalent to NNA, PNA, etc, that tells PARETO what constraints should be built. We can do that either explicitly, that is another tab in the Spreadsheet, or we can build the set automatically based on the tab `NodeCapacities` and apply that set in `def NetworkNodeCapacityRule(model, n, t)`. I personally like building the parameter automatically, but that's up for discussion @melody-shellman @drouvenm your feedback is appreciated. 
Also we need to add a config option for "Node Capacities" and default it to `False`. This option should be used to check if we need to build parameter **`model.p_sigma_NetworkNode`** and constraint **`model.NetworkCapacity`**

Andresj89 (changes requested):

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: That's a fair point, I agree it is better to avoid creating constraints that are redundant. One thought: we could have two columns in NodeCapacity (a binary if the node has a capacity constraint and capacity value). These can generate the parameters and the constraints are only created for the nodes with a "1" binary. 

I know a binary indicator may not be intuitive for all our users, but it would eliminate having an extra tab in our already ~67 tab document.

NienkeWagenaar: I like building the parameter automatically too. I think it's clear that if there's a value, it should be applied, and not if there's no value. I'll also set up a config option!

NienkeWagenaar: I've added a config option for including the capacity at all. The current setup will skip the constraint is no value is entered in the input sheet. I've added a not about this in the excel file too.

melody-shellman (approved):
Looks good!

Andresj89 (approved):
LGTM

PR #116, opened 2022-03-29
Fixing scaling factors

Andresj89:

It was reported that some of the configuration options for pipelines were resulting in the `v_C_Slack` having small negative numbers in the scaled model which after being propagated to the original model ended up with values in the order of millions.

- The scaling factors for all the slack variables, except for `v_C_Slack`, were changed to 1. This allows to reduce the scaling factor of `v_C_Slack` so that when the solution is propagated, any small amount does not have significant impact in the final solution
- NumericFocus is now included and set to 1

### Legal Acknowledgement

contribution:

   employer.


melody-shellman (commented):
Looks good! I tested the previous setup that was resulting in large negative slack values, and no longer see these.

melody-shellman (approved):
Looks good! I tested the previous setup that was resulting in large negative slack values, and no longer see these.

NienkeWagenaar (approved):
Looks good to me!

PR #117, opened 2022-04-06
Set slacks to zero

Andresj89:

This PR adds a method to set slacks to zero for the strategic model. The method checks if the termination condition of the solver is unfeasible, in that case the slacks are activated and the model is run again.

- The `run_strategic_model.py` has been simplified. It now contains call to the method `solve_model(model, options)` which receives a model object and a dictionary with the following options: 
```
options = {
    "deactivate_slacks": True,
    "scale_model": True,
    "scaling_factor": 1000000,
    "running_time": 60,
    "gap": 0,
    "water_quality" True}
```
- I modified `create_model()` slightly to "attach" `df_sets` and `df_parameters` to the model object so we don't have to pass those in as part of the `postprocess_water_quality_calculation(model, opt)`.
### Legal Acknowledgement

contribution:

   employer.


pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5265,6 +5268,111 @@ def _preprocess_data(model, _df_parameters):
...
+            TransformationFactory("core.scale_model").propagate_solution(
+                scaled_model, model
+            )
+
+    else:
+        print(
+            "WARNING: Model is infeasible. Adding Slack variables to avoid infeasibilities\n, \
+            however this is an indication that the input data should be revised"
+        )
+
```
NienkeWagenaar, line 5330: Can we move the unfixing to before the if-statement?

NienkeWagenaar (changes requested):

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -4498,7 +4501,7 @@ def CompletionsWaterDeliveriesRule(model, p, t):
     return model
 
 
-def water_quality(model, df_parameters, df_sets):
+def water_quality(model):
     # region Fix solved Strategic Model variables
     for var in model.component_objects(Var):
         for index in var:
```
melody-shellman, line 4507: We are still rounding small values to 0 when fixing variables prior to solving water quality. I quickly tested removing this rounding and the model is infeasible -- this is something I plan to address in a future PR.

melody-shellman (changes requested):
Overall looks good. I like this new restructuring of code, it is definitely cleaner. I ran the code with our four modes (capacity vs distance, input vs calculated) and all outputs looked good! Only change suggested is to remove df_parameters as an input to  **_preprocess_data (line 5212)**, similar to the other methods.

NienkeWagenaar (dismissed):

NienkeWagenaar:
I removed two other unused imports. Other than that it looks good!

melody-shellman (dismissed):
Looks good, thanks for implementing those suggestions!

melody-shellman (approved):
Looks good, runs as expected.

NienkeWagenaar (approved):
Looks great!

PR #135, opened 2022-06-07
Updated sankey visualization method to support generating sankeys for different regions

MAZamarripa:
… multiple regions

(Replace this with the issue # fixed or addressed in this PR as well as any other related PRs.  If
no issue or PRs exist, then add a brief statement on what changes this PR is bringing in.)

-
-

### Legal Acknowledgement

contribution:

   employer.

### DIFF
--- a/pareto/utilities/results.py
+++ b/pareto/utilities/results.py
@@ -633,9 +633,18 @@ def plot_sankey(input_data={}, args=None):
     if is_sections:
         for key, val in input_data["sections"].items():
             for i, v in enumerate(val):
-                v_index = label.index(v)
-                val[i] = v_index
+                try:
+                    v_index = label.index(v)
+                    val[i] = v_index
+                except:
+                    print(
+                        "WARNING: {0} does not have a value for every specified time period and may not be shown in the sankey diagram.".format(
+                            val[i]
+                        )
+                    )
+                    pass
 
+            val = [x for x in val if not isinstance(x, str)]
             df_section = df_updated[
                 (df_updated["source"].isin(val) | df_updated["destination"].isin(val))
             ]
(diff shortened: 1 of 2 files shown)


pareto/utilities/results.py
```diff
@@ -967,7 +1033,7 @@ def plot_bars(input_data, args):
         )
 
         # Update animation settings
-        fig.layout.updatemenus[0].buttons[0].args[1]["frame"]["duration"] = 200
+        fig.layout.updatemenus[0].buttons[0].args[1]["frame"]["duration"] = 800
```
melody-shellman, line 1036: When I run the tests locally, they run and complete within 23min. When the tests run on github they do not finish in 35 minutes. I'm not really sure why ...does changing these figure layout affect build time ?

melody-shellman (commented):
The tests are passing when running locally, but not on git. ... at least not within 35 min. Tests have typically run in <20min. Here are a few things that I can think of to reduce this time ....
-  I noticed that test_plot_scatter.py has 40 tests. Why do we need so many? Are all of these necessary?
- Do the changes to allow for input multi regions affect more than a few of these tests? Might this be why the testing time has increased drastically?
- Perhaps it's the "frame duration" setting that affects build time?

melody-shellman:
One more thing - it may also be good to add the output files of the pytest in the git ignore. I think it is just these files:
        Region 1_first_sankey.html
        Region 2_first_sankey.html
        first_bar.html
        first_sankey.html
        first_scatter.html

pareto/utilities/results.py
MAZamarripa: @lbianchi-lbl @dangunter can you check this issue? it could be the iplot() method.

melody-shellman:
This work will be completed in PR #149

PR #173, opened 2022-11-02
Adding Hydraulics module to the strategic model

nareshsusarla:

This PR adds the hydraulics module to the strategic model. The hydraulics module uses the Hazen Williams equation to compute the frictional head loss and the uses the energy balance equation to compute pressure at each node as a post optimization calculation.

- addition of hydraulics method that computes pressure at each node
- updating the run_strategic_model method to include config options for hydraulics method
- updating the excel file to include elevation data, a master set for locations, and initial pipeline diameter param
- Note on Input Sheets: 0 values will override default values in code. 0 values have been removed. InitialPipelineDiameters tab is used only for Hydraulics. This tab should only have values above the symmetry single diagonal. If there are values A->B and B->A this indicates two initial pipelines and is treated as a single pipeline with the sum of both pipeline diameters. 
- Note on JupyterNotebooks: content of notebooks has not changed with this PR, only formatting by Black .

### Legal Acknowledgement

contribution:

   employer.


melody-shellman:
PR #174  to be merged first !

ksbeattie:
@nareshsusarla is this something for a June release?

pareto/utilities/get_data.py
```diff
@@ -351,15 +352,18 @@ def get_display_units(input_sheet_name_list, user_units):
         "CST": "",
         "CCT": "",
         "CKT": "",
+        "Elevation": pyunits.foot,
```
tarnold17, line 355:
```suggestion
        "Elevation": "foot",
```
All of the other entries in this dictionary are strings, so I think this should probably be a string as well. (If you commit this change then you can also remove the line `from pyomo.environ import units as pyunits` from the file).

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -70,15 +71,18 @@
...
     "DesalinationTechnologies",
     "DesalinationSites",
     "TruckingTime",
     "CompletionsDemand",
     "PadRates",
     "FlowbackRates",
+    "WellPressure",
     "NodeCapacities",
     "InitialPipelineCapacity",
+    "InitialPipelineDiameters",
```
tarnold17, line 85: Is the onus on the user to ensure that the data entered in "InitialPipelineDiameters" is compatible with the data for "InitialPipelineCapacity"?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1783,7 +1869,7 @@ def create_model(df_sets, df_parameters, default={}):
                 from_units=model.user_units["currency_volume"],
                 to_units=model.model_units["currency_volume"],
             )
-            for key, value in model.df_parameters["PipelineOperationalCost"].items()
+            for key, value in _df_parameters.items()
```
tarnold17, line 1872: So, if I am reading the code correctly, if `config.hydraulics` is not `Hydraulics.post_process`, then all of the values for `model.p_pi_Pipeline` will be 0.0001 (see line 1762), and anything in the "PipelineOperationalCost" tab will be ignored. Is that the desired behavior? That doesn't seem like what we should want.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
     return model
 
 
+def operational_hydraulics(model, opt):
```
tarnold17, line 3491: Given that this is the strategic model, not the operational model, I think that the function name `operational_hydraulics` is potentially confusing and I suggest changing it to something else.

pareto/tests/test_strategic_model.py
tarnold17, line 1: We should have tests for all three of the hydraulics options - `Hydraulics.false`, `Hydraulics.post_process`, and `Hydraulics.co_optimize`.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17, line 3491: I believe this function will throw an exception if it is called on a model for which `config.hydraulics` is `Hydraulics.false`. I think it would be good to print a specific warning/error message in this telling the user not to call this function on a model with that config option.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+        for key in model.s_LLA:
+            mh.p_diameter_Pipeline_eff[key] = model.df_parameters[
+                "InitialPipelineDiameters"
+            ][key] + sum(
+                model.vb_y_Pipeline[key, d]
+                * model.df_parameters["PipelineDiameterValues"][d]
+                for d in model.s_D
+            )
+
+        # Compute Hazen Williams head
```
tarnold17, line 3600:
```suggestion
        # Compute Hazen-Williams head
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+        # The cloned model will be a MINLP formulation. The exisiting objective
+        # must be modified to add operating costs for pumps and valves.
+        mh = model.clone()
+        del mh.objective
+
+        mh.p_HW_material_factor_pipeline = Param(
+            initialize=130,
+            mutable=True,
+            units=pyunits.dimensionless,
+            doc="Pipeline material factor for Hazen Williams equation",
```
tarnold17, line 3686:
```suggestion
            doc="Pipeline material factor for Hazen-Williams equation",
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+            tmpdir="temp",
+            add_options=["gams_model.optfile=1;"],
+        )
+
+    return mh
+
+
+def _hazen_williams_head(mat_factor, length, diameter, flow):
+    """
+    Computes Hazen Williams (HW) head in a pipeline
```
tarnold17, line 3917:
```suggestion
    Computes Hazen-Williams (HW) head in a pipeline
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+            bounds=(mh.p_Min_AOP, None),
+            units=model.model_units["pressure"],
+            doc="Pressure at location l at time t in Pa",
+        )
+        for p in model.s_P:
+            for t in model.s_T:
+                if value(mh.p_WellPressure[p, t]) > 0:
+                    mh.v_Pressure[p, t].fix(mh.p_WellPressure[p, t])
+
+        mh.p_diameter_Pipeline_eff = Param(
```
tarnold17, line 3573: What does "eff" mean here?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+    ----------
+    mat_factor: pipeline material factor for HW equation
+    length : length of pipeline segment in m.
+    diameter : diameter of pipeline segment in m.
+    flow : water flow through the pipeline segment in m3/s.
+
+    Returns
+    -------
+    hw_friction_head : TYPE
+        DESCRIPTION.
```
tarnold17, line 3929: This documentation for the return value should be cleaned up.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+                * model.df_parameters["PipelineDiameterValues"][d]
+                for d in model.s_D
+            )
+
+        # Compute Hazen Williams head
+        for t0 in model.s_T:
+            for key in model.s_LLA:
+                if value(model.v_F_Piped[key, t0]) > 0.01:
+                    if value(mh.p_diameter_Pipeline_eff[key]) > 0.1:
+                        mh.p_HW_loss[key, t0] = _hazen_williams_head(
```
tarnold17, line 3605: The first three arguments passed to `_hazen_williams_head` are Pyomo expressions, but the fourth argument is a float. So, I think then `_hazen_williams_head` will return a Pyomo expression. Is that what you want? Or does it need to return a float instead? I am not sure here.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+        function of minmizing the cost is modified to include of the cost of
+        operating pumps and valves as a function of the deltaP.
+        """
+        # Clone the model to add hydraulics constraints
+        # The cloned model will be a MINLP formulation. The exisiting objective
+        # must be modified to add operating costs for pumps and valves.
+        mh = model.clone()
+        del mh.objective
+
+        mh.p_HW_material_factor_pipeline = Param(
```
tarnold17, line 3682: Beginning here, there's a good bit of code which is repeated from the `Hydraulics.post_process` section. Is there a way to consolidate/reformulate to avoid repeating a bunch of things?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+            units=pyunits.meter,
+            doc="Valve Head removed in the direction of flow, m",
+        )
+        mh.p_rhog = Param(
+            initialize=9.8 * 1000,
+            units=model.model_units["pressure"] / pyunits.meter,
+            doc="g (m/s2) * density of PW (assumed to be 1000 kg/m3)",
+        )
+        mh.p_Min_AOP = Param(
+            initialize=10 * 6895,
```
tarnold17, line 3532: I see `10` here and `14.7` a few liens below - should they match?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
```
tarnold17, line 3710: I see `10` here and `14.7` a few liens below - should they match?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -111,6 +122,21 @@ class RemovalEfficiencyMethod(Enum):
         }""",
     ),
 )
+CONFIG.declare(
+    "hydraulics",
+    ConfigValue(
+        default=Hydraulics.false,
+        domain=In(Hydraulics),
+        description="add hydraulics constraints",
+        doc="""Alternate pipeline capacity selection (calculated or input)
```
melody-shellman, line 131:
```suggestion
        doc="""Selection to include hydraulics
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1676,6 +1719,48 @@ def create_model(df_sets, df_parameters, default={}):
             units=model.model_units["pipe_cost_capacity"],
             doc="Pipeline construction/expansion capital cost for selected increment [currency/volume/time]",
         )
+
+    _df_parameters = {}
```
melody-shellman, line 1723: I suggest being more specific with this variable name. Something along the lines _elevation_adjusted_piping_cost?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1676,6 +1719,48 @@ def create_model(df_sets, df_parameters, default={}):
...
+                                )
+                                * elevation_delta
+                                / max_elevation_change
+                            )
+                        ),
+                    )
+    else:
+        for k1 in model.s_L:
+            for k2 in model.s_L:
+                _df_parameters[(k1, k2)] = 0.0001
```
melody-shellman, line 1762: Travis' comment at line 1872 is a good point. It looks like the input data is not being used for False and co optimize config settings. 
_df_parameters (or new variable name) should be set the same way we used to initialize the operating piping cost. 
{
            key: pyunits.convert_value(
                value,
                from_units=model.user_units["currency_volume"],
                to_units=model.model_units["currency_volume"],
            )
            for key, value in model.df_parameters["PipelineOperationalCost"].items()

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: This may be why the objective changes slightly in the tests. All piping costs may be set to the default value instead of reading from input. See my comment at 1762.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5987,4 +6521,14 @@ def solve_model(model, options=None):
         )
 
     results.write()
+    # ms.to_json(model, fname='optimal_strategic_model_solution_for_hydraulics.json')
+    # ms.from_json(model, fname='optimal_strategic_model_solution_for_hydraulics.json')
+    if model.config.hydraulics != Hydraulics.false:
+        hydraulics_model = operational_hydraulics(model, opt)
+
     return results
+    # results.write()
```
melody-shellman, line 6530: Remove extra commented code

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5987,4 +6521,14 @@ def solve_model(model, options=None):
         )
 
     results.write()
+    # ms.to_json(model, fname='optimal_strategic_model_solution_for_hydraulics.json')
```
melody-shellman, line 6524: Remove extra commented code.

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: Doesn't hurt to add a warning message, but the way the model is build now, this method is NOT called if Hydraulics config is false, so I don't thing it is essential. 
See line 6257:
```
if model.config.hydraulics != Hydraulics.false:
        hydraulics_model = operational_hydraulics(model, opt)
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
     return model
 
 
+def operational_hydraulics(model, opt):
+
+    if model.config.hydraulics == Hydraulics.post_process:
+        """
+        For the post process method, the pressure is computed in using
```
melody-shellman, line 3495:
```suggestion
        For the post process method, the pressure is computed using the
```

melody-shellman (changes requested):
Overall looking good. Some changes + questions about objective function changing. One more thing that I need to think on a bit more - units. We introduced some new units and use units directly in definitions as opposed to defining for the full model or for user. I will think on this a bit and offer my thoughts.

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: thanks, i agree. i updated it with `_df_parameters = model.df_parameters["PipelineOperationalCost"]`

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: i am not sure if i should rename it, because the variable is the same whether we use hydraulics or not

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -111,6 +122,21 @@ class RemovalEfficiencyMethod(Enum):
...
+    "hydraulics",
+    ConfigValue(
+        default=Hydraulics.false,
+        domain=In(Hydraulics),
+        description="add hydraulics constraints",
+        doc="""Alternate pipeline capacity selection (calculated or input)
+        ***default*** - False
+        **Valid Values:** - {
+        **Hydraulics.false** - does not add hydraulics feature
+        **Hydraulics.post_process** - adds economic parameters for flow based on elevation cahnges and computes pressures at each node post-optimization,
```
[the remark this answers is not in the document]
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: updated

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -104,7 +108,7 @@
     "PipelineCapexCapacityBased",
     "PipelineCapacityIncrements",
     "PipelineExpansionDistance",
-    "Hydraulics",
+    "PipelineExpansionDistance",
```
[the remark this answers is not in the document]
nareshsusarla: agree. removed it

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+        due to pumps and valves.
+        """
+        # Create a model and add hydraulics constraint to compute pressure at each node
+        mh = ConcreteModel()
+
+        mh.p_HW_material_factor_pipeline = Param(
+            initialize=130,
+            mutable=True,
+            units=pyunits.dimensionless,
+            doc="Pipeline material factor for Hazen Williams equation",
```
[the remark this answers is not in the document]
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: i agree. i would think the warning is probably not necessary

nareshsusarla: updated it to call the method pipeline_hydraulics

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: the objective values is restored

pareto/utilities/get_data.py
nareshsusarla: updated the units

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: I agree, however, the trade off is either repeat the definition of variables or check for the hydraulics config arg multiple times. I have now switched back to checking the config arg multiple times and build the model appropriately

pareto/strategic_water_management/run_strategic_model.py
nareshsusarla: I believe so. can explain offline

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: The numbers have now been moved to the spreadsheet and should be the same everywhere

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: The numbers have now been moved to the spreadsheet and should be the same everywhere

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3402,6 +3488,454 @@ def SeismicActivityExceptionRule(model, k, t):
...
+            initialize=0,
+            within=NonNegativeReals,
+            mutable=True,
+            units=pyunits.meter,
+            doc="Hazen-Williams Frictional loss, m",
+        )
+        for key in model.s_LLA:
+            mh.p_diameter_Pipeline_eff[key] = model.df_parameters[
+                "InitialPipelineDiameters"
+            ][key] + sum(
```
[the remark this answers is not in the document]
nareshsusarla: yes and it is now updated

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: The method is now updated for consistency. However, in either case, the method `_hazen_williams_head` only returns a float

pareto/tests/test_strategic_model.py
nareshsusarla: i have now added tests for post_process and co_optimize

pareto/utilities/get_data.py
```diff
@@ -20,6 +20,7 @@
 import pandas as pd
 import requests
 import numpy as np
+from pyomo.environ import units as pyunits
```
tarnold17, line 23:
```suggestion

```
This line is no longer needed and should be removed.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: @nareshsusarla this comment is definitely a nit pick, but it has not been addressed.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -34,9 +35,11 @@
     Suffix,
     TransformationFactory,
     value,
+    SolverFactory,
 )
 
 from pyomo.core.base.constraint import simple_constraint_rule
+from idaes.core.util import model_serializer as ms
```
tarnold17, line 42:
```suggestion

```
Pylance says this import is unused, so I recommend removing it.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -45,6 +48,8 @@
 
 from pareto.utilities.solvers import get_solver, set_timeout
 from pyomo.opt import TerminationCondition
+from idaes.core.util.model_statistics import degrees_of_freedom
+from pyomo.util.infeasible import log_infeasible_constraints, log_close_to_bounds
```
tarnold17, line 52:
```suggestion

```
More unused imports.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -98,6 +109,21 @@ class RemovalEfficiencyMethod(Enum):
...
     ),
 )
+CONFIG.declare(
+    "hydraulics",
+    ConfigValue(
+        default=Hydraulics.false,
+        domain=In(Hydraulics),
+        description="add hydraulics constraints",
+        doc="""option to include pipeline hydraulics in the model either during or post optimization
+        ***default*** - False
```
tarnold17, line 119:
```suggestion
        ***default*** - Hydraulics.false
```

pareto/strategic_water_management/run_strategic_model.py
tarnold17: Following up on this comment - should the data in the "InitialPipelineDiameters" tab only be in one direction (like for "InitialPipelineCapacity"), and not two directions? This screenshot is from the "InitialPipelineDiameters" tab in "strategic_toy_case_study.xlsx": 
image

I would think that e.g. you want to remove the entry for N02 -> N01, leaving instead just the entry for N01 -> N02, but I'm not sure.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -6014,4 +6447,71 @@ def solve_model(model, options=None):
         )
 
     results.write()
+
+    if model.config.hydraulics != Hydraulics.false:
+        mh = pipeline_hydraulics(model, opt)
```
tarnold17, line 6452:
```suggestion
        mh = pipeline_hydraulics(model)
```

pareto/tests/test_strategic_model.py
```diff
@@ -440,6 +448,78 @@ def _call_model_with_config(config_dict):
...
+            "objective": Objectives.cost,
+            "pipeline_cost": PipelineCost.capacity_based,
+            "pipeline_capacity": PipelineCapacity.input,
+            "hydraulics": Hydraulics.post_process,
+            "water_quality": WaterQuality.false,
+        }
+    )
+    solver = get_solver("cbc")
+    results = solver.solve(m, tee=False)
+    mh = pipeline_hydraulics(m, solver)
```
tarnold17, line 467:
```suggestion
    mh = pipeline_hydraulics(m)
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: I see - `pipeline_hydraulics()` (formerly `operational_hydraulics()`) is not meant to be called directly - rather, it is called within `solve_model()`, and it is only called there `if model.config.hydraulics != Hydraulics.false`.

Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3405,6 +3494,350 @@ def SeismicActivityExceptionRule(model, k, t):
...
+    )
+    mh.p_rhog = Param(
+        initialize=9.8 * 1000,
+        units=model.model_units["pressure"] / pyunits.meter,
+        doc="g (m/s2) * density of PW (assumed to be 1000 kg/m3)",
+    )
+    mh.p_PumpingCost = Param(
+        initialize=1.2,
+        doc="Cost of pumping water",
+    )
```
tarnold17, line 3555: Why 1.2? Should the user be allowed to modify/specify this parameter? As of now, it appears there isn't a way for the user to change it away from 1.2.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3405,6 +3494,350 @@ def SeismicActivityExceptionRule(model, k, t):
...
+    def MAOPressureRule(b, l1, t1):
+        constraint = b.v_Pressure[l1, t1] <= b.p_Max_AOP
+        return process_constraint(constraint)
+
+    mh.MAOPressure = Constraint(
+        model.s_L - model.s_P,
+        model.s_T,
+        rule=MAOPressureRule,
+        doc="Max allowable pressure rule",
+    )
```
tarnold17, line 3614: I don't see a corresponding constraint for min allowable pressure - is that something that needs to be added?

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -143,6 +148,7 @@
         "objective": Objectives.cost,
         "pipeline_cost": PipelineCost.distance_based,
         "pipeline_capacity": PipelineCapacity.input,
+        "hydraulics": Hydraulics.post_process,
```
tarnold17, line 151:
```suggestion
        "hydraulics": Hydraulics.false,
```
I think we want the default hydraulics setting to be `false`.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -6014,4 +6447,71 @@ def solve_model(model, options=None):
         )
 
     results.write()
+
+    if model.config.hydraulics != Hydraulics.false:
+        mh = pipeline_hydraulics(model, opt)
+
+        if model.config.hydraulics != Hydraulics.post_process:
```
tarnold17, line 6454:
```suggestion
        if model.config.hydraulics == Hydraulics.post_process:
```
As discussed on Slack.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -6014,4 +6447,71 @@ def solve_model(model, options=None):
...
+            results_2 = opt.solve(mh, tee=True)
+
+        elif model.config.hydraulics != Hydraulics.co_optimize:
```
tarnold17, line 6457:
```suggestion
        elif model.config.hydraulics == Hydraulics.co_optimize:
```
As discussed on Slack.

nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: done

pareto/strategic_water_management/run_strategic_model.py
nareshsusarla: done

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: the min allowable pressure is added as a bound but the max allowable pressure is added as a constraint to allow the flexibility to change the value during the analysis

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: removed

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: updated the text

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1756,6 +1788,63 @@ def create_model(df_sets, df_parameters, default={}):
...
+        )
+    if model.config.hydraulics == Hydraulics.post_process:
+        # In the post-process based hydraulics model an economics-based penalty is added
+        # to moderate flow through pipelines based on the elevation change.
+        max_elevation = max(
+            [abs(val) for val in model.df_parameters["Elevation"].values()]
+        )
+        min_elevation = min(
+            [abs(val) for val in model.df_parameters["Elevation"].values()]
+        )
```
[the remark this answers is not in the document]
nareshsusarla: updated to remove abs

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3405,6 +3494,350 @@ def SeismicActivityExceptionRule(model, k, t):
     return model
 
 
+def pipeline_hydraulics(model, opt):
```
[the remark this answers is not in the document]
nareshsusarla: removed

pareto/tests/test_strategic_model.py
nareshsusarla: changed the test implementation

pareto/tests/test_strategic_model.py
```diff
@@ -440,6 +448,78 @@ def _call_model_with_config(config_dict):
...
+            "objective": Objectives.cost,
+            "pipeline_cost": PipelineCost.capacity_based,
+            "pipeline_capacity": PipelineCapacity.input,
+            "hydraulics": Hydraulics.co_optimize,
+            "water_quality": WaterQuality.false,
+        }
+    )
+    solver = get_solver("cbc")
+    results = solver.solve(m, tee=False)
+    mh = pipeline_hydraulics(m, solver)
```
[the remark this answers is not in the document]
nareshsusarla: changed the test implementation

lbianchi-lbl:
@nareshsusarla we discussed the possibility that some changes related to review comments might not have been pushed yet to this PR - can you check this? Also, let me know if you encounter issues with Git/GitHub with this.

lbianchi-lbl:
@nareshsusarla will add the documentation to this PR for the July release.

tarnold17:
#241 is now merged which will cause a merge conflict with this PR over the file `operational_generic_case_study.xlsx`, but according to @melody-shellman you should be able to just take the version of this file that is now on main and you will be good to go.

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: This parameter is now replaced by more granular and mutable parameters like electricity price, and efficiencies

pareto/strategic_water_management/strategic_produced_water_optimization.py
nareshsusarla: in addition to the aforementioned comment, the MaxAOP is not added as a bound because it is only enforced for some nodes. For example, the production pads and completion pads are not subjected to this constraint.

pareto/utilities/results.py
```diff
@@ -558,6 +559,105 @@ def generate_report(
...
+                            "Time",
+                            "Pressure at a Location",
+                        )
+                    ],
+                    "hydraulics.v_HW_loss_dict": [
+                        (
+                            "Location",
+                            "Location",
+                            "Time",
+                            "Hazen Williams head loss",
```
tarnold17, line 649:
```suggestion
                            "Hazen-Williams head loss",
```

pareto/strategic_water_management/run_strategic_model.py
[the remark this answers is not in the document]
tarnold17: Resolving.

pareto/tests/test_strategic_model.py
```diff
@@ -638,7 +644,7 @@ def test_run_reduced_strategic_model(build_reduced_strategic_model):
     assert results.solver.status == pyo.SolverStatus.ok
     assert degrees_of_freedom(m) == 11789
     # solutions obtained from running the reduced generic case study
-    assert pytest.approx(89201.666, abs=1e-1) == pyo.value(m.v_Z)
+    assert pytest.approx(88199.598, abs=1e-1) == pyo.value(m.v_Z)
```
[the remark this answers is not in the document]
tarnold17: @nareshsusarla this is still showing a slight change in the objective function value - do you have an explanation?

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: I agree that a more descriptive variable name would be helpful here. I think I would probably just go with something like `pipeline_operational_cost`. The variable is used regardless of whether or not hydraulics are being used, but in any scenario, it's being used to hold the pipeline operational cost values.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Looks good to me now. Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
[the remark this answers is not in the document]
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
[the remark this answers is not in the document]
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/tests/test_strategic_model.py
tarnold17: Resolving.

pareto/tests/test_strategic_model.py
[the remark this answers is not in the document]
tarnold17: Resolving.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Thanks for the explanation; resolving.

pareto/strategic_water_management/run_strategic_model.py
tarnold17: @nareshsusarla I would still like an explanation here - perhaps best for us to discuss on a call though.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Resolving.

pareto/strategic_water_management/run_strategic_model.py
tarnold17: We discussed this on a call and agreed that the structure of the "InitialPipelineDiameters" tab should be changed as I suggested above, to remove symmetrical entries (similar to the "InitialPipelineCapacity" tab).

pareto/examples/strategic_model_general_demo/strategic_model_general_demo.ipynb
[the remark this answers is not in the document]
nareshsusarla: Yes that is correct

pareto/tests/test_strategic_model.py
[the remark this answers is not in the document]
melody-shellman: Note: change in objective is expected as previous versions of small_strategic_model contain no values for PipelineOperationalCosts, meaning the default value is being used (0.01 instead of 0.0001 $/bbl)

pareto/tests/test_strategic_model.py
```diff
@@ -440,6 +448,131 @@ def _call_model_with_config(config_dict):
...
+    assert results.solver.termination_condition == pyo.TerminationCondition.optimal
+    assert results.solver.status == pyo.SolverStatus.ok
+    assert degrees_of_freedom(m) == 11789
+    # solutions obtained from running the reduced generic case study
+    assert pytest.approx(88199.598, abs=1e-1) == pyo.value(m.v_Z)
+    assert pytest.approx(26.067, abs=1e-1) == pyo.value(m.hydraulics.v_Z_HydrualicsCost)
+    assert pytest.approx(24, abs=1e-1) == pyo.value(
+        sum(m.hydraulics.vb_Y_Pump[key] for key in m.s_LLA)
+    )
+    # feasibility test is not working with Cbc so, disabling for now until a fix is found
```
ElmiraShamlou, line 528: The infeasibility issue returned by CBC primarily stems from the violation of the max allowable pressure constraint (mh.MAOPressure). This, I believe, is predominantly due to the scaling challenge associated with the model's pressure unit being in Pascal, which inherently presents large numerical values.

To mitigate this, we can consider two potential solutions:

1) Change the pressure unit from Pascal to bar, adjust the initial value of rhog from 9.8 * 1000 to 9.8 * 1000 * 1e-5, and revise the PumpCostRule as follows:
def PumpCostRule(b, l1, l2):
    constraint = b.v_PumpCost[l1, l2] == mh.p_PumpFixedCost * mh.vb_Y_Pump[l1, l2] + ((mh.p_ElectricityCost / 3.6e6) * mh.p_rhog * 1e5 * sum(b.v_PumpHead[l1, l2, t] * pyunits.convert_value(value(model.v_F_Piped[l1, l2, t]), from_units=model.model_units["volume_time"], to_units=pyunits.meter**3 / pyunits.h) for t in model.s_T))
    return process_constraint(constraint)

2) Alternatively, we can multiply both sides of the mh.MAOPressure and nodePressureRule constraints by 1e-5 to effectively address the scaling issue.

I believe adopting either of these approaches should also contribute to resolving issue #268

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -3405,6 +3487,445 @@ def SeismicActivityExceptionRule(model, k, t):
...
+                    == b.v_Pressure[l2, t1]
+                    + model.p_Elevation[l2] * mh.p_rhog
+                    + b.p_HW_loss[l1, l2, t1] * mh.p_rhog
+                    - b.v_PumpHead[l1, l2, t1] * mh.p_rhog
+                    + b.v_ValveHead[l1, l2, t1] * mh.p_rhog
+                )
+                return process_constraint(constraint)
+            else:
+                return Constraint.Skip
+
```
ElmiraShamlou, line 3748: I'm curious if this pressure rule can accurately apply to storage tanks, particularly given the dynamic nature of the tank's head level. Additionally, the significant reduction of fluid velocity within the tank seems to be a factor not fully accounted for (by either the Hazen-Williams equation or the valve head loss calculations).

ElmiraShamlou: Suggestion: Alternatively, we can consider adding a constraint for storage tanks, where we equate tank head to the total of tank water level and tank elevation, to better represent the static nature of water in the tank. Even if we were to assume a constant tank level, incorporating this new constraint can help represent the change in velocity within the tank. This act like forcing a valve head loss prior to tank.

nareshsusarla: Thanks for the comment. The strategic model doesn't really have an explicit storage tank included in it. The head calculation would require detailed geometry of the storage tank / pond (recall we only see capacities). Now, coming to your point of managing pressure. The node pressure can be viewed as the pressure at which water enters the pipeline system so, the required pressure corrections (if any) will be done at the node level. This is similar to assuming a lumped node for tank and a pump/valve. While you suggest is a nice to have fidelity for the model, it is a detail not necessary at the strategic level, at least for now.

docs/model_library/Hydraulics/index.rst
```diff
@@ -0,0 +1,204 @@
...
+
+:math:`\textcolor{green}{\eta^{Pump}}` =                        Efficiency of the pump
+
+:math:`\textcolor{green}{\eta^{Motor}}` =                        Efficiency of the motor for pump
+
+:math:`\textcolor{green}{\xi^{Minimum}}` =                        Minimum allowable operating pressure in a pipeline
+
+:math:`\textcolor{green}{\xi^{Maximum}}` =                        Maximum allowable operating pressure in a pipeline (MAOP)
+
+:math:`\textcolor{green}{D_{l,\tilde{l}}^{Existing}}` =                        Diameter of an exisiting pipeline
```
ElmiraShamlou, line 53:
```suggestion
:math:`\textcolor{green}{D_{l,\tilde{l}}^{Existing}}` =                        Diameter of an existing pipeline
```

docs/model_library/Hydraulics/index.rst
```diff
@@ -0,0 +1,204 @@
...
+a.	Hydraulics.false: This option allows the user to skip the hydraulics computations in the PARETO model.
+b.	Hydraulics.post_process: In this method, the basic PARETO strategic model is solved as step 1, and then using the optimal flows and network design, the hydraulics block containing constraints for pressure balances and losses is solved as step 2. In this case, as only the hydraulics model block solved for the objective of minimizing cost, the optimal values for variables included in the main strategic model and obtained from step 1 remain unaffected.
+c.	Hydraulics.co_optimize: In this method, the hydraulics model block is solved together with the strategic model. However, as the flow and diameter are variables in the strategic model, the addition of hydraulics block makes the model a mixed integer nonlinear programming (MINLP) model. In order to solve this MINLP model, the strategic model without the hydraulics constraints is solved as step 1 to determine a good initial state for all variables and constraints.
```
ElmiraShamlou, line 13:
```suggestion

a)	Hydraulics.false: This option allows the user to skip the hydraulics computations in the PARETO model.

b)	Hydraulics.post_process: In this method, the basic PARETO strategic model is solved as step 1, and then using the optimal flows and network design, the hydraulics block containing constraints for pressure balances and losses is solved as step 2. In this case, as only the hydraulics model block solved for the objective of minimizing cost, the optimal values for variables included in the main strategic model and obtained from step 1 remain unaffected.

c)	Hydraulics.co_optimize: In this method, the hydraulics model block is solved together with the strategic model. However, as the flow and diameter are variables in the strategic model, the addition of hydraulics block makes the model a mixed integer nonlinear programming (MINLP) model. In order to solve this MINLP model, the strategic model without the hydraulics constraints is solved as step 1 to determine a good initial state for all variables and constraints.
 
```

melody-shellman (dismissed):
Approved - great work!

docs/model_library/Hydraulics/index.rst
```diff
@@ -0,0 +1,208 @@
...
++--------------------------------------------------------+
+| :ref:`mathematical_model_formulation`                  |
++--------------------------------------------------------+
+
+.. _hydraulics_model_mathematical_notation:
+
+Hydraulics Model Mathematical Notation
+-------------------------------------------
+
+Similar to the strategic model, following color coding has been implemented in describing the model notation. All \textcolor{green}{input parameters are in green} and \textcolor{red}{model variables are in red}.
```
ElmiraShamlou, line 35: This part of the document does not render correctly in the built HTML

pareto/strategic_water_management/strategic_produced_water_optimization.py
ElmiraShamlou: Thank you for your response. As we have discussed offline, it is beneficial to consider introducing an additional constraint to more accurately represent the storage pressure and the difference in storage inlet and outlet pressure. One potential way is to add:
 Pstorage(t) = H(t) * rho.g
 where dH/dt = 1/A * dQ/dt ,  
(H: the storage tank water level plus the storage location elevation, Q: the net flow to and out of the storage, A: storage cross sectional area).
However, we may want to set this potential improvement aside for the moment and consider revisiting it with a later PR.

ElmiraShamlou: Thank you for your response. As we have discussed offline, it is beneficial to consider introducing an additional constraint to more accurately represent the storage pressure and the difference in storage inlet and outlet pressure. One potential way is to add:
 Pstorage(t) = H(t) * rho.g
 where dH/dt = 1/A * dQ/dt ,  
(H: the storage tank water level plus the storage location elevation, Q: the net flow to and out of the storage, A: storage cross sectional area).
However, we may want to set this potential improvement aside for the moment and consider revisiting it with a later PR.

docs/model_library/Hydraulics/index.rst
ElmiraShamlou, line 35:
```suggestion
Similar to the strategic model, following color coding has been implemented in describing the model notation. All input :math:`\textcolor{green}{parameters}` are in :math:`\textcolor{green}{green}` and all model :math:`\textcolor{red}{variables}` are in :math:`\textcolor{red}{red}`.
```

ElmiraShamlou (approved):
Absolutely great job! Approved

melody-shellman:
Issue #268 resolved with commits to this PR. Note: it was NOT a CBC issue turns out, it was a scaling issue which has been resolved by changing model units for pressure in the hydraulics module.

PR #184, opened 2022-12-22
Strategic model - add convert default value for p_pi_Trucking from user units to model units

### LINKED ISSUE #182: Strategic model - add pyunits.convert_value() in default value for p_pi_Trucking parameter
The default value of the trucking hourly cost parameter, `p_pi_Trucking`, is missing a call to `pyunits.convert_value()` to convert from user units to model units.

tarnold17:

In the strategic model, added a call to `pyunits.convert_value` to convert the default value for trucking hourly cost (`p_pi_Trucking`) from user units to model units.

Resolves #182 

### Legal Acknowledgement

contribution:

   employer.

### DIFF
--- a/pareto/strategic_water_management/strategic_produced_water_optimization.py
+++ b/pareto/strategic_water_management/strategic_produced_water_optimization.py
@@ -1732,7 +1732,11 @@ def create_model(df_sets, df_parameters, default={}):
     # be confusing
     model.p_pi_Trucking = Param(
         model.s_L,
-        default=max(model.df_parameters["TruckingHourlyCost"].values()) * 100
+        default=pyunits.convert_value(
+            max(model.df_parameters["TruckingHourlyCost"].values()) * 100,
+            from_units=model.user_units["currency"],
+            to_units=model.model_units["currency"],
+        )
         if model.df_parameters["TruckingHourlyCost"]
         else pyunits.convert_value(
             15000,
--- a/pareto/strategic_water_management/strategic_produced_water_optimization.py
+++ b/pareto/strategic_water_management/strategic_produced_water_optimization.py
@@ -1727,30 +1727,27 @@ def create_model(df_sets, df_parameters, default={}):
         units=model.model_units["currency_volume"],
         doc="Pipeline operational cost [currency/volume]",
     )
+    TruckingHourlyCost_convert_to_model = {
+        key: pyunits.convert_value(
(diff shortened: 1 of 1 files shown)


melody-shellman (changes requested):
Looks good! This was a good catch.  looking at some of our other parameter default values, it looks like these may need correction as well:

- [ ] model.p_lambda_Pipeline
- [ ] model.p_pi_Disposal
- [ ] model.p_pi_Reuse
- [ ] model.p_pi_Sourcing

Can we apply the same correction to these parameters, as well?

tarnold17:
> Looks good! This was a good catch. looking at some of our other parameter default values, it looks like these may need correction as well:
> 
> * [ ]  model.p_lambda_Pipeline
> * [ ]  model.p_pi_Disposal
> * [ ]  model.p_pi_Reuse
> * [ ]  model.p_pi_Sourcing
> 
> Can we apply the same correction to these parameters, as well?

@melody-shellman the four parameters listed here already have the default values converted from user units to model units. For example, the default value for `p_lambda_Pipeline` is `max(PipelineExpansionDistance_convert_to_model.values()) * 100`, where `PipelineExpansionDistance_convert_to_model` is constructed using a dictionary comprehension to  convert all of the popeline expansion distances from user units to model units.

melody-shellman:
Oh I see! Good point. Agreed  that the parameters I listed no longer need adjustment.

One suggestion instead, is it possible to have the formats for `p_pi_Trucking` match that of those I listed for consistency? More specifically, can we create the TruckingHourlyTrucking_convert_to_model dictionary that applies the conversion? With this dictionary, we can call the max without conversion function and we can initialize with this dictionary instead of a conversion.

tarnold17:
> Oh I see! Good point. Agreed that the parameters I listed no longer need adjustment.
> 
> One suggestion instead, is it possible to have the formats for `p_pi_Trucking` match that of those I listed for consistency? More specifically, can we create the TruckingHourlyTrucking_convert_to_model dictionary that applies the conversion? With this dictionary, we can call the max without conversion function and we can initialize with this dictionary instead of a conversion.

@melody-shellman no problem - addressed in latest commit.

melody-shellman (approved):
Looks good! Results are the same as main for treatment demo case, which is reasonable.

ElmiraShamlou (approved):
Looks good.

PR #186, opened 2022-12-23
Clean up small strategic case study and turn off default scaling for strategic model

### LINKED ISSUE #185: Clean up small strategic case study input file and schematic
The following actions need to be applied to the small strategic case study input file (`small_strategic_case_study.xlsx`) and to the corresponding schematic (https://app.box.com/file/1062715916286).

- [x] In the NNA tab of the input file, the N19 - N22 and N26 - N27 arcs should be input both ways instead of just one.
- [x] In the schematic, remove the arc between N28 and the end user option (the beneficial reuse node O01 is included in the input file, but it is not connected to N28 because we do not yet support any arcs to or from beneficial reuse sites).
- [x] Remove the line between N23 and CP04 from the schematic (this arc is not included in the CNA tab of the input file).
- [x] In the input file, remove the arc between N10 and N12 in the NNA tab (this is not supposed to be a valid arc - the schematic shows that PP05 is between N10 and N12).
- [x] Remove all references to "FF" and "HDH" from the input file. These are desalination technologies, but both of the treatment sites in the small strategic case study are for clean brine treatment, so it is impossible for FF or HDH to ever get picked. Removing any mention of FF and HDH from the input file will help to avoid confusing users.
- [x] Add a capacity increment of 0 as an option for treatment technology CB.
- [x] Remove disposal nodes K03 and K05 from the input file and the schematic (and then rename K04 to K03). K03 and K05 have zero initial disposal capacity, and the small strategic case study does not allow disposal capacity expansion to occur, so there is no purpose to keeping these disposal sites in the network.
- [x] Update the NKA tab of the input file to include the N01 - K01 arc that is depicted in the schematic.
- [x] Add upper limits in PadOffloadingCapacity for CP02 - CP04 to match the limit for CP01.
- [x] In tab PipelineOperationalCost, the only nonzero entries are between CP01-CP01, CP02-CP02, CP03-CP03, and CP04-CP04. These are not valid arcs, so remove them from the tab.
- [x] In tab TruckingHourlyCost, add trucking costs for F03-F08.
- [x] In tab TruckingTime, add trucking times for PP04-PP15.
- [x] Storage site S01 should be removed from it's current location in the input file and schematic. Instead, S01 should be connected to R01, CP01, and CP02. Similarly, a second storage site, S02, should be added which is connected to R02, CP03, and CP04.
- [x] Resolve inconsistencies between the InitialPipelineCapacity tab in the input file and the schematic. There are numerous examples where the schematic shows an arc with zero initial pipeline capacity, but the input file shows nonzero initial pipeline capacity. We want to update the schematic to match what is currently in the input file.
- [x] General cleanup and formatting of the input file.

tarnold17:

Resolves #185 

Turns off scaling by default for strategic model

### Legal Acknowledgement

contribution:

   employer.


tarnold17:
I had an issue where one of the tests in `test_strategic_model.py` was failing for me locally, but not on GitHub. @melody-shellman helped investigate, and she found that the cause of the problem was that scaling was causing Gurobi to detect the model with the water quality extension as infeasible. The reason that I was only noticing the failure locally is that GitHub uses CBC for tests, whereas locally, Gurobi is used.

We decided that the best course of action is to change the strategic model such that it is not scaled by default. Scaling is now often unneeded for the strategic model since units were added several months ago - the conversion from user units to model units provides a similar effect as scaling. The scaling option remains part of the strategic model, and it can still be used if needed, but it doesn't need to be used by default.

melody-shellman (approved):
Thank you for these detailed changes on the input! This PR definitely improves the quality of our small_strategic_case_study.

I did take a look at the results and how they differ from main. There is more trucking, more sourced water, more disposal, and less reuse with these changes. Usually, I'd raise a flag to look into this more, but I don't think this is too concerning because the results didn't make much sense before anyways. 

I will be using this version of input as a starting point for developing a more realistic case study for our meeting in Feb.

ElmiraShamlou (approved):
lgtm

Issue #188, opened 2023-01-05
UI does not currently display units; add dictionary to map tab name to unit type that can be utilized by UI


melody-shellman:
UI does not currently display units. The fix is to add dictionary to map tab name to unit type that can be utilized by UI. To complete:

- [x] implement for strategic model
- [x] implement for operational model
- [x] add tests to strategic and operational to unsure there is a unit for each input tab

PR #189, opened 2023-01-05
UI Display Units

melody-shellman:

Resolves issue #188 

- Created new function `get_display_units()` in `get_data` to output UI display units
- Added tests to strategic and operational model to ensure that new tabs have defined output units
- removed unused tab DisposalPipeCapEx from input sheet and `get_data()` reads

@MichaelPesce Note that this implementation is slightly different than what we discussed. Previously we discussed mapping each tab to the unit type referenced in the "Units" tab. The issue with this, is units are often compounded types from the "Units" tab. Our most common unit, for example, is bbl/week, which is made up of multiple types "volume"/"week". To simplify this, I wrote a function that takes a list of sheet names and the "Units" tab and exports the tab to be displayed. One other note -users could possible change units in the units tab, in this case, the units should probably be updated throughout, so this function may need to be called whenever the units tab is updated. 
Let me know if have any questions!
### Legal Acknowledgement

contribution:

   employer.


tarnold17 (approved):

MichaelPesce (approved):
looks good to me

PR #190, opened 2023-01-06
Nm UI small case study

melody-shellman:

- This PR creates a new configuration option for

-
-

### Legal Acknowledgement

contribution:

   employer.

### DIFF
--- a/pareto/strategic_water_management/run_strategic_model.py
+++ b/pareto/strategic_water_management/run_strategic_model.py
@@ -13,6 +13,7 @@
 
 from pareto.strategic_water_management.strategic_produced_water_optimization import (
     WaterQuality,
+    BuildUnits,
     create_model,
     Objectives,
     solve_model,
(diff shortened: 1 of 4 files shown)


pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -131,6 +133,7 @@
         "pipeline_capacity": PipelineCapacity.input,
         "node_capacity": True,
         "water_quality": WaterQuality.false,
+        "build_units": BuildUnits.user_units,
```
tarnold17, line 136: I wonder, should the default value here be `scaled_units` instead of `user_units`? I understand that the benefit of using `user_units` is reduced runtime for the script, but using `scaled_units` is the current default behavior for `run_strategic_model.py` so I'm not sure that we should change that. Personally, I anticipate that my preference will generally be to use the `scaled_units` option, despite the additional runtime.

tarnold17 (changes requested):
The changes look pretty good to me for the most part, but the tests are not currently passing.

pareto/strategic_water_management/run_strategic_model.py
melody-shellman: Agreed! this is addressed in recent commits.

tarnold17 (approved):

MichaelPesce (approved):

PR #199, opened 2023-01-23
Add case study documentation

tarnold17:

Add initial documentation for case studies.

- Rename case study input files to be shorter

### Legal Acknowledgement

contribution:

   employer.


drouvenm (approved):
I like this whole page a lot @tarnold17! The table in particular, should be incredibly helpful to new PARETO users (industrial or academic). Coupled with the schematics (including the updated labels), I imagine everyone on the team will benefit from this documentation. Well done!

melody-shellman (approved):
Looks great! An excellent first pass. If time allows, I thought of another row that would be useful to add - 
Compatible Solvers: CBC (open source), Gurobi (commercial)
I believe all but the treatment demo are compatible with both. Treatment demo requires Gurobi

drouvenm:
@melody-shellman On the note of solver selections, someone needs to show me where/how I can do that for the strategic model. It's so straightforward for the operational model but I didn't even find that option for the strategic run file...

tarnold17:
> @melody-shellman On the note of solver selections, someone needs to show me where/how I can do that for the strategic model. It's so straightforward for the operational model but I didn't even find that option for the strategic run file...

@drouvenm you can select the solver for the strategic model by passing a "solver" option to the `solve_model` function. You are right to notice that we don't do that in `run_strategic_model.py`, so running this script uses the default behavior, which is to use Gurobi, and if it can't find Gurobi, then CBC instead.

tarnold17:
> Looks great! An excellent first pass. If time allows, I thought of another row that would be useful to add - Compatible Solvers: CBC (open source), Gurobi (commercial) I believe all but the treatment demo are compatible with both. Treatment demo requires Gurobi

@melody-shellman I think the issue is not really one of "compatibility", but rather, that CBC just isn't a good enough solver to solve the strategic treatment demo in a reasonable amount of time. However, I think it's possible that there are optimizations/improvements we could make to the strategic model that would allow it to be solved relatively quickly with CBC. So I decided to take your suggestion, but I worded it as Gurobi being "recommended" for the treatment demo.

melody-shellman:
> > Looks great! An excellent first pass. If time allows, I thought of another row that would be useful to add - Compatible Solvers: CBC (open source), Gurobi (commercial) I believe all but the treatment demo are compatible with both. Treatment demo requires Gurobi
> 
> @melody-shellman I think the issue is not really one of "compatibility", but rather, that CBC just isn't a good enough solver to solve the strategic treatment demo in a reasonable amount of time. However, I think it's possible that there are optimizations/improvements we could make to the strategic model that would allow it to be solved relatively quickly with CBC. So I decided to take your suggestion, but I worded it as Gurobi being "recommended" for the treatment demo.

Agreed, the note you have in the latest commit I think is good!

melody-shellman (approved):
lgtm!

PR #203, opened 2023-01-30
Feasibility check

apd-ypuranik:

When a solver terminates with a status other than Optimal or Infeasible (TimeLimit, IterationLimit), it is unclear whether the solution returned is feasible. Knowing whether to accept an intermediate solution is helpful, since we are likely to hit TimeLimits with large models and/or open source solvers.

This PR implements a utility to check whether a solution is feasible or not to help in this situation. 

- Implement a method to check if a solution is feasible (Heavily inspired by: https://github.com/Pyomo/pyomo/blob/141baf028213236c08711105bb2fbb024aa28a66/pyomo/util/infeasible.py)
- Shows how to use the method when running strategic or operational models

### Legal Acknowledgement

contribution:

   employer.


apd-ypuranik:
Some issues raised during a live review session:

1. Will we also descend and check sub model components? A: Addressed with the use of method (model.component_data_objects(ctype=*, descend_into=True)) which recursively descends into and loops over all model variables or constraints as the case may be
2. Will units be a factor? A: No. The lower/upper and variable values are all attached to the same object and follow the same unit. Different variables could be in different scales and we could eventually consider using different tolerances for different variables, but for now, this is consistent with what any MIP solver may do.
3. What if variable values are None? None values are not expected as a replacement for 0. If a solver returns a solution, we should have variable values populated.

tarnold17 (approved):

melody-shellman (approved):
Great! Works as expected for water quality and with CBC with low runtime.

PR #206, opened 2023-01-30
add reduced treatment demo case study

melody-shellman:
- Add new case study strategic_Permian_demo.xlsx
- Add images for case study and results for scenarios

### Legal Acknowledgement

contribution:

   employer.


MichaelPesce (dismissed):

tarnold17 (changes requested):
Can you adjust the line breaks in the documentation table in the Permian demo column so that it has a similar width to the other columns?

I recommend not including spaces in filenames - e.g., change "Ppipeline Shutdown.png" to "pipline_shutdown.png".

docs/case_studies/index.rst
```diff
@@ -91,43 +108,56 @@ Comparison table
...
    * - **Completions pad**
+     - No
      - No
      - Yes
      - No
      - Yes
    * - **Treatment sites**
+     - 5 |br|
+       Non-desalination sites: R02, R04, R05 |br|
+       Desalination sites: R03, R06 |br|
```
tarnold17, line 158: Why isn't R01 listed as a desal site? It is included in the schematic.

melody-shellman:
> Can you adjust the line breaks in the documentation table in the Permian demo column so that it has a similar width to the other columns?
> 
> I recommend not including spaces in filenames - e.g., change "Ppipeline Shutdown.png" to "pipline_shutdown.png".

The naming is intentional for a new feature in the UI. If the naming matches the name of a scenario, it will autopopulate in the UI.

tarnold17 (dismissed):
Looks good to me - one last final question I have though - do you think we should add a test to make sure that the model builds correctly using `strategic_permian_demo.xlsx`? We have that test for all of the other case studies in the repo, so my inclination is that we should add it for this one as well.

MichaelPesce (approved):

MichaelPesce (approved):

tarnold17 (approved):

Issue #213, opened 2023-02-14
Investigate strategic Permian demo returning infeasible for water quality post processing with Gurobi 


tarnold17:
When solving with Gurobi (with scaling either on or off), and `WaterQuality.post_process` selected for the `water_quality` build option, the strategic Permian demo reports an infeasible model for the post processing calculations.

An interesting note - if one solves this problem instead with CBC with scaling turned on, there are no problems.

melody-shellman:
- only an issue with Gurobi
- infeasible solution returned from Gurobi (not just the feasibility checker)

melody-shellman:
I believe this is an issue with scaling. Here's what supports that:

- The first step of the post_processing of water quality solves the model without quality and fixes the solution.
- While fixing all variables, if a value is < 0.0000001, it is rounded to 0. Gurobi seems to have some interesting behavior in that optimal solutions are sometimes not actually optimal (tolerances being a factor) 
- I tested changing this threshold to 0.00001, and the problem now solves. 

Other notes:
- Running with scaling increases the size of these 'zero' values from (R04 to S04 T37) 10^-5 to 10^-2

tarnold17:
@melody-shellman good to know. Perhaps we should discuss this with the rest of the team at tomorrow afternoon's treatment meeting. It sounds like the solution is some form of "improve scaling", but do you have any ideas more specifically than that how we should proceed?

melody-shellman:
Related to #209, will likely be resolved with the same solution

melody-shellman:
PR #214 resolved this issue with a correction to the post_process water quality treatment center constraints.

PR #214, opened 2023-02-17
Treatment plant component removal efficiency

ElmiraShamlou:

Incorporate component removal efficiency to the model

- Add treatment center removal efficiency to the model and modify balances accordingly.
- Rename technology set b/B to wt/WT to prevent confusion and attribute errors when working with methods in the water_quality block. 
- Create variable treatment feed to maintain treatment feed quality in the results and for future use in capital cost estimation based on feed flow. 
- Create a new intermediate set for residual water quality tracking. Update component set w/W to qc/QC. 
- Include removal efficiency in case_studies, tests, results, and get_data files. Update the strategic model library document to reflect these changes.

### Legal Acknowledgement

contribution:

   employer.


pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1337,11 +1345,20 @@ def create_model(df_sets, df_parameters, default={}):
...
         default=1.0,
         initialize=model.df_parameters["TreatmentEfficiency"],
         doc="Treatment efficiency [%]",
     )
+
+    model.p_epsilon_Treatment_Removal = Param(
+        model.s_R,
+        model.s_WT,
+        model.s_QC,
+        default=1.0,
```
tarnold17, line 1358: A minor point, but I think I'd recommend using 0 as the default parameter value for treatment removal.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1800,6 +1817,15 @@ def create_model(df_sets, df_parameters, default={}):
...
         doc="Big-M flow parameter [volume/time]",
     )
+
+    model.p_M_Concentration = Param(
+        default=100,
+        units=model.model_units["concentration"],
+        doc="Big-M concentration parameter [concentration]",
+    )
+
+    model.p_M_Flow_Conc = model.p_M_Flow * model.p_M_Concentration
```
tarnold17, line 1827: I'm not sure this is a good way to define this big-M parameter. If we need a big-M parameter with units of flow * concentration, I would use another call to `Param()` to crate it.

docs/model_library/strategic_water_management/index.rst
```diff
@@ -1451,51 +1451,51 @@ For :math:`t > 1`:
         + \sum_{k \in K | (s, k) \in SKT}\textcolor{purple}{F_{s,k,t}^{Trucked}}
         + \textcolor{purple}{F_{s,t}^{StorageEvaporationStream}})
 
-**Treatment Site Water Quality** :math:`\forall \textcolor{blue}{r \in R}, \textcolor{blue}{w \in W}, \textcolor{blue}{t \in T}`
+**Treatment Site Water Quality** :math:`\forall \textcolor{blue}{r \in R}, \textcolor{blue}{qc \in QC}, \textcolor{blue}{t \in T}`
 
 The water quality at treatment sites is dependent on the flow rates into the treatment site and the water quality of the flows. Even mixing is assumed, so all outgoing flows have the same water quality. The treatment process does not affect water quality.
```
tarnold17, line 1456: This part of the documentation should be updated - the model has changed, and the treatment process now does affect water quality.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5513,102 +5565,149 @@ def StorageSiteWaterQualityRule(b, s, w, t):
...
-    def TreatedWaterWaterQualityRule(b, r, w, t):
-        constraint = (b.parent_block().v_F_TreatedWater[r, t]) == b.v_Q[
-            r + treatment_intermediate_label, w, t
-        ] * (
-            sum(
-                b.parent_block().v_F_Piped[r, p, t]
-                for p in b.parent_block().s_CP
-                if b.parent_block().p_RCA[r, p]
+
+    def TreatedWaterQualityLHSRule(b, r, wt, qc, t):
```
tarnold17, line 5629: I see that you have rewritten this constraint to use a big-M formulation, but it doesn't seem like the documentation has been updated to reflect this change in the model. Can you please update the documentation so that it shows the new form of the constraints?

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: One more minor point - by our naming conventions, this variable should be model.p_epsilon_TreatmentRemoval (no underscore between Treatment and Removal). Can this also be updated? (e.g. model.p_sigma_NetworkNode)

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5239,30 +5279,40 @@ def water_quality(model):
         doc="Completions Pad Intermediate Flows",
     )
     # Create a set for water quality tracked at the intermediate node between treatment facility and treated water end points
-    treatment_intermediate_label = "-PostTreatmentIntermediateNode"
+    treated_water_label = "-PostTreatmentTreatedWaterNode"
     model.df_sets["TreatedWaterIntermediateNodes"] = [
```
melody-shellman, line 5283: I think the renaming you have here makes sense. Should we also rename the set "TreatedWaterIntermediateNodes" to "TreatedWaterNode"?

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: Agreed that the documentation needs updating for these new constraints / adjustment to WaterQualityRule. Updates needed in the general model and water quality sections.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5513,102 +5565,149 @@ def StorageSiteWaterQualityRule(b, s, w, t):
...
+    model.quality.TreatmentWaterQualityLHS = Constraint(
+        model.s_R,
+        model.s_WT,
+        model.s_QC,
+        model.s_T,
+        rule=TreatedWaterQualityLHSRule,
+        doc="Treatment water quality",
+    )
+
+    def TreatedWaterQualityRHSRule(b, r, wt, qc, t):
```
melody-shellman, line 5657: A question here - in the strategic model we have constraints for ResidualWaterLHS and ResidualWaterLHS, but in the water quality we look at it from a treated water perspective instead of residual water. Is there a reason for this difference? To me, it seems that is it mathematically the same, but just curious because it's a bit confusing that it doesn't match,

melody-shellman (commented):
Small changes inn the strategic_toy_case_study.xlsx file: 

- the "RemovalEfficiency" tab is unformatted and not categorized right. It should be categorized after "TreatmentEfficiency" like the other case studies. 

- Additionally the PadWaterQuality, StorageInitialWaterQuality, 

- and PadStorageInitialWaterQuallity titles do not display correctly. They are written correctly, but just need a "reset". I was able to correct it by opening the formula and hitting enter. 

Other notes:

- The TreatmentFeed variable cleans up some constraint to make it more readable - love it!

- As discussed, waiting for these small changes and update to discrete water quality model for a final review, but generally looks really good!

melody-shellman:
One more thing - what file are you using for testing? Are the results looking ok? I tried testing with the toy case study and these are the quality results:
I would have expected it to match the removal efficiencies. Am I wrong here?

image

ElmiraShamlou:
> One more thing - what file are you using for testing? Are the results looking ok? I tried testing with the toy case study and these are the quality results for T13: I would have expected it to match the removal efficiencies. Am I wrong here?
> 
@melody-shellman Your results from the toy case accurately reflect the removal efficiency. It's important to note that removal efficiency is based on the overall load of components (flow times concentration), not just the concentration of the components. As a result, calculating the concentration of the residual and treated water streams directly from the feed water quality and removal efficiency is not sufficient. You need to take both treatment efficiency (recovery) and removal efficiency into account when evaluating the quality of the treated water and residual stream.
To validate your results, you can use the two following equations:

Q(TreatedWater) = Q(TreatmentFeed) x (1 - RemovalEff) / TreatmentEff

Q(Residual) = Q(TreatmentFeed) x RemovalEff / (1 - TreatmentEff)

By substituting the treatment feed quality, removal efficiency, and treatment efficiency values into these two formulas, you can obtain the exact same number as your displayed result from the toy case study.

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: Corrected these!

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: done!

pareto/strategic_water_management/strategic_produced_water_optimization.py
ElmiraShamlou: In water treatment, the removal efficiency is typically defined as the ratio of the reduction in concentration/load of the contaminant in treated water to the concentration/load of the contaminant in the feed water. This method accounts for any losses of the contaminant/water during the treatment process that are not necessarily going into the residual water.

It is therefore better to express the removal efficiency constraint in terms of treated water and, depending on whether we have material losses, write the treatment balance constraint. If there are no losses, the material balance in the current formulation of strategic model is valid otherwise it should be changed. The same rational can be applied for the treatment efficiency formulation to account for any possible water losses, depending on the technology considered.
However, I did not change the already existing strategic model part for treatment efficiency as there is already a TreatedWater rule and to minimize the changes in this PR. It is important to note that removal efficiency and treatment efficiency are independent constraints that do not need to match.

melody-shellman: okay, that makes sense to me. I'll resolve this conversation.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -1337,11 +1345,20 @@ def create_model(df_sets, df_parameters, default={}):
...
     )
+
+    model.p_epsilon_TreatmentRemoval = Param(
+        model.s_R,
+        model.s_WT,
+        model.s_QC,
+        default=0,
+        initialize=model.df_parameters["RemovalEfficiency"],
+        doc="Removal efficiency [%]",
+    )
```
tarnold17, line 1361:
```suggestion
    model.p_epsilon_TreatmentRemoval = Param(
        model.s_R,
        model.s_WT,
        model.s_QC,
        default=0,
        initialize=model.df_parameters["RemovalEfficiency"],
        mutable=True,
        doc="Removal efficiency [%]",
    )
```

docs/model_library/strategic_water_management/index.rst
melody-shellman: Updated!

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: Updated!

pareto/strategic_water_management/strategic_produced_water_optimization.py
ElmiraShamlou: @tarnold17 I agree that the big M for the bilinear term of F*Q can be easily defined using only one parameter. However, the value of big M required for this term is defined with the units used. Because mass is conserved, mass balances in water treatment systems are commonly written as mass fraction times mass flow. Mass fraction is always between 0 and 1, which makes estimating the magnitude of the big M for mass fraction and flow times concentration intuitive. However, the current formulation of PARETO employs volume balances rather than mass balances, with concentration and flow units of kg/l and bbl/week. In adjusting the big M for the bilinear term F*Q, the current units are not intuitive. As a result, using a separate big M for concentration (kg/l) makes it easier for anyone working with PARETO to change it with any unit change. Furthermore, defining a big M for concentration may become necessary when setting upper bounds on concentration variables in future MINLP formulations with quality in bilinear terms.

tarnold17: So, `model.p_M_Flow` and `model.p_M_Concentration` are both Pyomo `Param`s - specifically, the have the type `pyomo.core.base.param.ScalarParam`. However, when you multiply these variables together, the result that you get is no longer a `Param`, it's an expression. Specifically, it has type `pyomo.core.expr.numeric_expr.NPV_ProductExpression`. This seems like bad practice to me. which is why I suggest using a call to `Param()` to create `model.p_M_Flow_Conc`.

pareto/strategic_water_management/strategic_produced_water_optimization.py
ElmiraShamlou: Thank you for raising a valid point about the efficiency of using an expr for the model.p_M_Flow_Conc parameter. To address your concern and avoid any confusion with the bilinear term unit, I have defined model.p_M_Flow_Conc as a separate parameter that is initialized using model.p_M_Flow and model.p_M_Concentration. This approach also allows us to keep model.p_M_Concentration as the big-M parameter for concentration and use it in our MINLP.

pareto/strategic_water_management/strategic_produced_water_optimization.py
ElmiraShamlou: @tarnold17 Thank you for raising a valid point about the efficiency of using an expression for the model.p_M_Flow_Conc parameter. To address your concern and avoid any confusion with the bilinear term unit, I have defined model.p_M_Flow_Conc as a separate parameter that is initialized using model.p_M_Flow and model.p_M_Concentration. This approach also allows us to keep model.p_M_Concentration as the big-M parameter for concentration and use it in our MINLP.

melody-shellman:
An FYI for those following this PR. This PR is ALMOST there. I am working with @ElmiraShamlou  to finalize treatment efficiency and treatment removal efficiency values in our case studies.

docs/model_library/strategic_water_management/index.rst
```diff
@@ -1517,69 +1548,69 @@ The water quality at the completions pad intermediate node is dependent on the f
...
-        + \sum_{s \in S | (s, p) \in SCA}\textcolor{purple}{F_{s,p,t}^{Piped}} \cdot \textcolor{red}{Q_{s,w,t}}
+    \sum_{n \in N | (n, p) \in NCA}\textcolor{purple}{F_{n,p,t}^{Piped}} \cdot \textcolor{red}{Q_{n,qc,t}}
+        + \sum_{\tilde{p} \in P | (\tilde{p}, p) \in PCA}\textcolor{purple}{F_{\tilde{p},p,t}^{Piped}} \cdot \textcolor{green}{\nu_{\tilde{p},qc,[t]}}
+        + \sum_{s \in S | (s, p) \in SCA}\textcolor{purple}{F_{s,p,t}^{Piped}} \cdot \textcolor{red}{Q_{s,qc,t}}
 
-        + \sum_{\tilde{p} \in P | (\tilde{p}, p) \in CCA}\textcolor{purple}{F_{\tilde{p},p,t}^{Piped}} \cdot \textcolor{green}{\nu_{\tilde{p},w,[t]}}
-        + \sum_{r \in R | (r, p) \in RCA}\textcolor{purple}{F_{r,p,t}^{Piped}} \cdot \textcolor{red}{Q_{r,w,t}}
-        + \sum_{f \in F | (f, p) \in FCA}\textcolor{purple}{F_{f,p,t}^{Sourced}} \cdot \textcolor{red}{Q_{f,w,t}}
+        + \sum_{\tilde{p} \in P | (\tilde{p}, p) \in CCA}\textcolor{purple}{F_{\tilde{p},p,t}^{Piped}} \cdot \textcolor{green}{\nu_{\tilde{p},qc,[t]}}
+        + \sum_{r \in R | (r, p) \in RCA}\textcolor{purple}{F_{r,p,t}^{Piped}} \cdot \textcolor{red}{Q_{r^{TreatedWaterNodes},qc,t}}
```
tarnold17, line 1556: Why does this line have `Q_{r^{TreatedWaterNodes},qc,t}` instead of just `Q_{r,qc,t}`?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -5690,49 +5794,49 @@ def BeneficialReuseWaterQuality(b, o, w, t):
...
             for s in b.parent_block().s_S
             if b.parent_block().p_SCA[s, p]
         ) + sum(
-            b.parent_block().v_F_Piped[p_tilde, p, t] * b.v_Q[p_tilde, w, t]
+            b.parent_block().v_F_Piped[p_tilde, p, t] * b.v_Q[p_tilde, qc, t]
             for p_tilde in b.parent_block().s_CP
             if b.parent_block().p_CCA[p_tilde, p]
         ) + sum(
-            b.parent_block().v_F_Piped[r, p, t] * b.v_Q[r, w, t]
+            b.parent_block().v_F_Piped[r, p, t] * b.v_Q[r + treated_water_label, qc, t]
```
tarnold17, line 5815: This is essentially the same question as I put in the documentation, but why use `treated_water_label` here?

tarnold17 (changes requested):
This PR is very close to being ready. I have a request for a small change to the documentation, and a few questions, and beyond that per @melody-shellman there are some changes that still need to be made to the input files and then pushed. But on the whole, this is very nice work.

docs/model_library/strategic_water_management/index.rst
melody-shellman: At most nodes, there is a single water quality associated with the output water at that node. However, at treatment sites there is the quality associated with the mixed inlet (R),  treated water ( R-PostTreatmentTreatedWaterNode) and quality associated with the residual ( R-PostTreatmentResidualWaterNode). They are technically from the same node R, but to differentiate the qualities we append the tags. We have previously used the same logic for completions pad + completions pad storage sites. 
See lines here for their definitions: https://github.com/project-pareto/project-pareto/blob/84050a74dcdbd638b1b655056ad937fc94bf6b78/pareto/strategic_water_management/strategic_produced_water_optimization.py#L5265

pareto/strategic_water_management/strategic_produced_water_optimization.py
ElmiraShamlou: @tarnold17 To expand on @melody-shellman response, it's worth noting that the treated water (with quality of  Q[r + treated_water_label, qc, t]) - as opposed to the treatment feed (with quality of Q[r, qc, t])) - is sent to completion pads for reuse. This information is also reflected in the Treated Water Balance constraint:

    def TreatedWaterRule(model, r, t):
        constraint = (
            model.v_F_TreatedWater[r, t]
            == sum(model.v_F_Piped[r, p, t] for p in model.s_CP if model.p_RCA[r, p])
            + sum(model.v_F_Piped[r, s, t] for s in model.s_S if model.p_RSA[r, s])
            + model.v_F_DesalinatedWater[r, t]
        )
        return process_constraint(constraint)

docs/model_library/strategic_water_management/index.rst
```diff
@@ -1451,51 +1463,69 @@ For :math:`t > 1`:
         + \sum_{k \in K | (s, k) \in SKT}\textcolor{purple}{F_{s,k,t}^{Trucked}}
         + \textcolor{purple}{F_{s,t}^{StorageEvaporationStream}})
 
-**Treatment Site Water Quality** :math:`\forall \textcolor{blue}{r \in R}, \textcolor{blue}{w \in W}, \textcolor{blue}{t \in T}`
+**Treatment Feed Water Quality** :math:`\forall \textcolor{blue}{r \in R}, \textcolor{blue}{qc \in QC}, \textcolor{blue}{t \in T}`
 
-The water quality at treatment sites is dependent on the flow rates into the treatment site and the water quality of the flows. Even mixing is assumed, so all outgoing flows have the same water quality. The treatment process does not affect water quality.
+The water quality at treatment sites is dependent on the flow rates into the treatment site and the water quality of the flows. Even mixing is assumed, so all outgoing flows have the same water quality.
```
tarnold17, line 1468:
```suggestion
The water quality at treatment sites is dependent on the rate and quality of the feed stream(s) into the treatment site. Even mixing is assumed in calculating the quality of the combined feed stream.
```

docs/model_library/strategic_water_management/index.rst
```diff
@@ -1451,51 +1463,69 @@ For :math:`t > 1`:
         + \sum_{k \in K | (s, k) \in SKT}\textcolor{purple}{F_{s,k,t}^{Trucked}}
         + \textcolor{purple}{F_{s,t}^{StorageEvaporationStream}})
 
-**Treatment Site Water Quality** :math:`\forall \textcolor{blue}{r \in R}, \textcolor{blue}{w \in W}, \textcolor{blue}{t \in T}`
+**Treatment Feed Water Quality** :math:`\forall \textcolor{blue}{r \in R}, \textcolor{blue}{qc \in QC}, \textcolor{blue}{t \in T}`
 
-The water quality at treatment sites is dependent on the flow rates into the treatment site and the water quality of the flows. Even mixing is assumed, so all outgoing flows have the same water quality. The treatment process does not affect water quality.
+The water quality at treatment sites is dependent on the rate and quality of the feed stream(s) into the treatment site. Even mixing is assumed in calculating the quality of the combined feed stream.
```
tarnold17, line 1468:
```suggestion
The water quality at treatment sites is dependent on the flow rates and qualities of the feed streams into the treatment site. Even mixing is assumed in calculating the quality of the combined feed stream.
```
Sorry to do this again - I should have been more careful the first time I rewrote this sentence.

tarnold17 (approved):
This is ready now, as far as I'm concerned. Great work @ElmiraShamlou !

melody-shellman (approved):
Treatment efficiency  and removal values look good! I tested and results look good (aside from the infeasibilities we are already aware of)!

melody-shellman:
Issue #213 was resolved with this PR

Issue #215, opened 2023-02-21
Allow some parameters in the strategic model to be mutable


tarnold17:
We should add a config option to the strategic model to allow certain parameters to be created as mutable in order to more easily enable sensitivity analysis.

tarnold17:
It turns out that `Param` objects in Pyomo are necessarily mutable if they have units. In other words, it is impossible to create an immutable `Param` with units. A lot of the `Param` objects in the strategic model have units, and therefore are mutable. Nevertheless, there are a few parameters in the model which do not have units, and could possibly be useful to alter in a sensitivity analysis. For those such parameters, we should change them to be created as mutable.

PR #222, opened 2023-03-02
Strategic model - make some Params mutable

tarnold17:

Addresses #215 

### Legal Acknowledgement

contribution:

   employer.


melody-shellman (commented):
A couple of notes I learned from this PR:

- Params are inherently mutable if units are defined for it
- Pyomo will throw an error if a mutable Param is used in an if statement to build a constraint

Other notes:

- Not yet approving, as I believe PR #214 introduces a variable `model.p_epsilon_TreatmentRemoval` that should be mutable. [UPDATE] implemented with the PR itself
- To discuss: should our arcs be mutable? [UPDATE] it is not possible to make the arcs mutable. The way the model is constructed changes depending on the arcs.

melody-shellman:
@MAZamarripa and @nareshsusarla do we also want to make our arcs (CNA, NNA, etc.) mutable? One example where we have seen this is the "pipeline shutdown" case in which we remove and arc from CNA. Not sure if we have many other examples of that, however.

melody-shellman (approved):
All concerns have been addressed! Looks good!

tarnold17:
> @MAZamarripa and @nareshsusarla do we also want to make our arcs (CNA, NNA, etc.) mutable? One example where we have seen this is the "pipeline shutdown" case in which we remove and arc from CNA. Not sure if we have many other examples of that, however.

For the record: we discussed this as a team and concluded that arc Params need to remain immutable. If someone wants to change the structure of the arcs in their network, this necessitates rebuilding the entire model.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -688,16 +688,19 @@ def create_model(df_sets, df_parameters, default={}):
     model.p_chi_OutsideCompletionsPad = Param(
         model.s_CP,
         initialize=model.df_parameters["CompletionsPadOutsideSystem"],
+        mutable=False,  # this Param must be immutable since it is used in if statements
```
nareshsusarla, line 691: Minor: do we need to keep this in the options? I am guessing we are not specifying the mutable option for all the parameters.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -688,16 +688,19 @@ def create_model(df_sets, df_parameters, default={}):
...
     model.p_chi_DesalinationTechnology = Param(
         model.s_WT,
         initialize=model.df_parameters["DesalinationTechnologies"],
+        mutable=False,  # this Param must be immutable since it is used in if statements
         doc="Binary parameter designating the treatment technologies for Desalination",
     )
     model.p_chi_DesalinationSites = Param(
         model.s_R,
         initialize=model.df_parameters["DesalinationSites"],
+        mutable=False,  # this Param must be immutable since it is used in if statements
```
nareshsusarla, line 703: same as before

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -688,16 +688,19 @@ def create_model(df_sets, df_parameters, default={}):
     model.p_chi_OutsideCompletionsPad = Param(
         model.s_CP,
         initialize=model.df_parameters["CompletionsPadOutsideSystem"],
+        mutable=False,  # this Param must be immutable since it is used in if statements
         doc="Binary parameter designating the Completion Pads that are outside the system",
     )
     model.p_chi_DesalinationTechnology = Param(
         model.s_WT,
         initialize=model.df_parameters["DesalinationTechnologies"],
+        mutable=False,  # this Param must be immutable since it is used in if statements
```
nareshsusarla, line 697: same as above

nareshsusarla (commented):
If the parameters are not mutable, can we have have the options consistent across the model?

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: The reason I explicitly included `mutable=False` here is more because it gave me a good place to include the comment which explains why these Params have to be immutable.

tarnold17: I ended up just removing the whole line.

nareshsusarla (approved):
lgtm

melody-shellman (approved):
looks good!

PR #228, opened 2023-03-27
add config argument for removal efficiency calculation method

ElmiraShamlou:

This pull request adds a new `config` argument for removal efficiency calculation using concentration-based and load-based approaches. 

- Config argument added
- Case studies updated according to concentration_based removal efficiency
- Tests updated

### Legal Acknowledgement

contribution:

   employer.


pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -137,7 +138,8 @@
         "pipeline_cost": PipelineCost.distance_based,
         "pipeline_capacity": PipelineCapacity.input,
         "node_capacity": True,
-        "water_quality": WaterQuality.false,
+        "water_quality": WaterQuality.post_process,
```
tarnold17, line 141: What is your rationale for changing the default for `"water_quality"` from `false` to `post_process`?

tarnold17:
At first glance, this looks great to me! A few quick comments:

1) Documentation updates should be a part of this PR - i.e., the documentation should explain both of the options for the removal efficiency calculations.
2) I think you should add a test to `test_strategic_model.py` to test the strategic model with the non-default option for `removal_efficiency_method`.

pareto/strategic_water_management/run_strategic_model.py
melody-shellman: Good question. I would think this should stay the same.

ElmiraShamlou: I made a change to test the treatment removal efficiency argument, but I forgot to revert the change before pushing. I updated it again to set the argument back to false.

melody-shellman (commented):
Minor comments! Tested with both configs options with toy case study and works as expected.

tarnold17 (approved):
Looks good to me - great work!

melody-shellman (approved):
Looks great! Thanks for adding extra notes/docs to the input sheet and run file. 
I confirmed expected results with post-process on toy case study.

PR #254, opened 2023-06-13
added tests for bounding functions & utilities

Tominapa:

Adds bounding functions and utilities. Tests implemented for all.

Note for future iterations of the bounding utility:

- Bounding for v_F_DiscreteTrucked requires changes to strategic model to be implemented: https://github.com/project-pareto/project-pareto/pull/254#discussion_r1260083135
- Bounding for v_Q  (p_Q_UB) is working, but can be improved: https://github.com/project-pareto/project-pareto/pull/254#discussion_r1255832089
- Remove redundant free() and unfix() references

### Legal Acknowledgement

contribution:

   employer.


lbianchi-lbl:
This is also ready to merge, pending reviews.

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+
+###############################################################################
+### Variable bounding manager functions
+def VariableBounds(model):
+    model = Bound_v_F_Piped(model)
+    if model.config.water_quality is WaterQuality.post_process:
+        model = Bound_v_Q_PP(model)
+    elif model.config.water_quality is WaterQuality.discrete:
+        model = Bound_v_F_DiscretePiped(model)
+        # model = Bound_v_F_Trucked(model) # TODO: resolve unit problem with p_max_number_of_trucks - it must be week^-1 to make the units work out
```
melody-shellman, line 35: Noticed this TODO. Is it still something that is needed?

pareto/utilities/model_modifications.py
melody-shellman, line 1: These are nifty and I could see being quite useful!

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+        default=None,
+        mutable=True,
+        # initialize=p_Q_UB_init,
+        units=model.model_units["concentration"],
+    )
+    # Assign upper bounds to variables
+    for l in model.quality.s_WQL:
+        for w in model.s_QC:
+            for t in model.s_T:
+                # Individual quality bounds (NOTE: HAVE SHOWN, THIS DOES NOT WORK)
```
melody-shellman, line 114: What does the "NOTE" here mean?

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+# The Regents of the University of California, through Lawrence Berkeley National Laboratory, et al.
+# All rights reserved.
+#
+# NOTICE. This Software was developed under funding from the U.S. Department of Energy and the U.S.
+# Government consequently retains certain rights. As such, the U.S. Government has been granted for
+# itself and others acting on its behalf a paid-up, nonexclusive, irrevocable, worldwide license in
+# the Software to reproduce, distribute copies to the public, prepare derivative works, and perform
+# publicly and display publicly, and to permit others to do so.
+#####################################################################################################
+# Title: Supporting functions for both Operational and Strategic models
```
melody-shellman, line 13: I think it would be helpful to add a note here that this method adds bounds to the following variables: v_F_Piped, v_F_Trucked, quality.v_Q, and v_F_DiscretePiped with a note on why others variables are not included.

melody-shellman (commented):
Looking good! I have a few comments/suggestions proposed.

pareto/utilities/bounding_functions.py
Tominapa: This will require a change in the corresponding PARETO parameter in the strategic model code. I built this in for future reference. At this point, I would rather keep this as-is and raise this in a future update.

pareto/utilities/bounding_functions.py
Tominapa: I tested the commented line that follows on 115 where the bound is derived from the p_Q_UB parameter. The logic that is used to create that parameter is not correct, resulting in a bound that is too tight, and the model solution will change with that bound enforced. I kept it there for two reasons: one, so it is available for testing should the same idea be tried in future, and two, as a record that the approach did not function properly. The current bound is fairly loose, and there is room for improvement in future versions.

pareto/utilities/bounding_functions.py
Tominapa: We could absolutely include other variables in future versions. The objective was to develop bounds for these three, being the variables involved in nonlinear functions in the MINLP formulation of the strategic model.

pareto/utilities/bounding_functions.py
tarnold17: @Tominapa that seems reasonable to me, but I think you might as well comment out all of the code for `model.quality.p_Q_UB` (lines 101-109) for now, since it is currently unused.

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+    model.v_C_Slack.fix(0)
+    model.v_S_FracDemand.fix(0)
+    model.v_S_Production.fix(0)
+    model.v_S_Flowback.fix(0)
+    model.v_S_PipelineCapacity.fix(0)
+    model.v_S_StorageCapacity.fix(0)
+    model.v_S_DisposalCapacity.fix(0)
+    model.v_S_TreatmentCapacity.fix(0)
+    model.v_S_ReuseCapacity.fix(0)
+    return None
```
tarnold17, line 60:
```suggestion
```
Minor, but you can remove this line if you want. Python returns `None` by default if a `return` statement is not explicitly included.

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+                index_var = var if index is None else var[index]
+                # unfix binary variables and unbound the continuous variables
+                if index_var.domain is Binary:
+                    index_var.unfix()
+                    index_var.free()
+                else:
+                    index_var.unfix()
+                    index_var.setlb(0)
+                    index_var.setub(None)
+    return None
```
tarnold17, line 46:
```suggestion
```
Same comment here.

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+                if not is_in_time_period:
+                    continue
+                index_var = var if index is None else var[index]
+                # unfix binary variables and unbound the continuous variables
+                if index_var.domain is Binary:
+                    index_var.unfix()
+                    index_var.free()
+                else:
+                    index_var.unfix()
+                    index_var.setlb(0)
```
tarnold17, line 44: Why are you setting the lower bound to zero here instead of `None`? Do we have any variables that could conceivably take on negative values?

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+                        if i in model.s_T and i not in time_period:
+                            is_in_time_period = False
+                            break
+                if not is_in_time_period:
+                    continue
+                index_var = var if index is None else var[index]
+                # unfix binary variables and unbound the continuous variables
+                if index_var.domain is Binary:
+                    index_var.unfix()
+                    index_var.free()
```
tarnold17, line 41: What does the `free()` function do?

pareto/tests/test_utilities.py
```diff
@@ -0,0 +1,337 @@
...
+    RemovalEfficiencyMethod,
+)
+from pareto.utilities.get_data import get_data
+from pareto.utilities.results import (
+    generate_report,
+    PrintValues,
+    OutputUnits,
+    is_feasible,
+    nostdout,
+)
```
tarnold17, line 33:
```suggestion
```
These imports are all unused and can be removed.

pareto/tests/test_utilities.py
```diff
@@ -0,0 +1,337 @@
...
+from pareto.utilities.bounding_functions import VariableBounds
+from pareto.utilities.model_modifications import free_variables
+from pareto.utilities.model_modifications import deactivate_slacks
+
+# import pytest
+import pytest
+
+
+############################
+def build_strategic_model(config_dict):
```
tarnold17, line 48: Two things:

1) Is there a reason you aren't using `@pytest.fixture(scope="module")` like we do in `test_strategic_model.py`?
2) This function doesn't just build the strategic model, it also solves it. So I recommend changing the name of the function to reflect that fact.

tarnold17 (commented):
I don't see any problems, but I do have quite a few questions.

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+    model = Bound_v_F_Piped(model)
+    if model.config.water_quality is WaterQuality.post_process:
+        model = Bound_v_Q_PP(model)
+    elif model.config.water_quality is WaterQuality.discrete:
+        model = Bound_v_F_DiscretePiped(model)
+        # model = Bound_v_F_Trucked(model) # TODO: resolve unit problem with p_max_number_of_trucks - it must be week^-1 to make the units work out
+    elif model.config.water_quality is WaterQuality.false:
+        pass
+    else:
+        pass
```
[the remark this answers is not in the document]
Tominapa: I did this in case any other options become available in future. This is more verbose, but also more future-proof.

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+        mutable=True,
+        within=Any,
+        initialize=p_F_Piped_UB_init,
+        units=model.model_units["volume_time"],
+        doc="Maximum pipeline capacity between nodes [volume_time]",
+    )
+    # Assign upper bounds to variables; note that using a bounds rule would require redefining the variable
+    for key in model.s_LLA:
+        for t in model.s_T:
+            model.v_F_Piped[key, t].setub(model.p_F_Piped_UB[key])
```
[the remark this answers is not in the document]
Tominapa: Solvers treat bounds and constraints differently. For MINLP formulations, having explicit bounds will be helpful.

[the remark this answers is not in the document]
Tominapa: Even if they are redundant.

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+        for w in model.s_QC:
+            for t in model.s_T:
+                # Individual quality bounds (NOTE: HAVE SHOWN, THIS DOES NOT WORK)
+                # model.quality.v_Q[l, w, t].setub(model.quality.p_Q_UB[l, w])
+                # Highest single quality bound
+                model.quality.v_Q[l, w, t].setub(quality_highest_bound)
+    return model
+
+
+def TryOrNone(par, ind):
```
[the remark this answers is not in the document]
Tominapa: I built this function to handle parameters that might not be defined on every index. At present, it isn't used, but it was used in earlier versions that had different logic.

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+#
+# NOTICE. This Software was developed under funding from the U.S. Department of Energy and the U.S.
+# Government consequently retains certain rights. As such, the U.S. Government has been granted for
+# itself and others acting on its behalf a paid-up, nonexclusive, irrevocable, worldwide license in
+# the Software to reproduce, distribute copies to the public, prepare derivative works, and perform
+# publicly and display publicly, and to permit others to do so.
+#####################################################################################################
+
+###--- Imports ---###
+from importlib import resources
```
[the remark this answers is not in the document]
Tominapa: Probably. It was probably copied in with other code.

pareto/utilities/bounding_functions.py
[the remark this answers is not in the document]
tarnold17: > Solvers treat bounds and constraints differently.

I did not realize this fact. Thanks for explaining!

pareto/utilities/bounding_functions.py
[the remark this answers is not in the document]
tarnold17: @Tominapa Do you expect it will be needed later? If we have no reason to believe that this function will be needed later, then I'd say just get rid of it.

pareto/tests/test_utilities.py
Tominapa: Addressed

pareto/tests/test_utilities.py
Tominapa: 1. I don't know what that first part is for. It can't be important though, or someone would have mentioned it earlier.
2. I need the strategic model with its quality block. If I could get it *without* solving the model, I would. The naming serves its purpose, in that respect.

pareto/utilities/bounding_functions.py
Tominapa: Added

pareto/utilities/bounding_functions.py
Tominapa: The parameter p_max_number_of_trucks will need to be available within the primary model block, or else this bound can only be added to specific subtypes. Moreover, the units on the parameter are [unitless] but I have a strong argument for making the units for p_max_number_of_trucks [week^-1]. I think this will need to be a follow-up PR once this first version is merged.

pareto/utilities/bounding_functions.py
Tominapa: Commented

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+    except:
+        val = None
+    finally:
+        return val
+
+
+def Bound_v_F_DiscretePiped(model):
+    # initialization function for parameter
+    def p_F_DiscretePiped_UB_init(model, l, l_tilde):
+        # set default value to pieline capacity plus maximum expansion capacity
```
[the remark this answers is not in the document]
Tominapa: Corrected

pareto/utilities/model_modifications.py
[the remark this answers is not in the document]
Tominapa: Removed.

pareto/utilities/model_modifications.py
Tominapa: The free() method unfixes the fixed value of a variable.

pareto/utilities/model_modifications.py
Tominapa: I like the explicit "return None" It's readable.

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,161 @@
...
+        model.s_L,
+        model.s_L,
+        default=None,
+        mutable=True,
+        within=Any,
+        initialize=p_F_Piped_UB_init,
+        units=model.model_units["volume_time"],
+        doc="Maximum pipeline capacity between nodes [volume_time]",
+    )
+    # Assign upper bounds to variables; note that using a bounds rule would require redefining the variable
```
nareshsusarla, line 71: What does it mean "using bounds rule would require redefining the variable"

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+from pyomo.environ import (
+    Var,
+    Binary,
+)
+
+###--- Functions ---###
+# free variables function
+def free_variables(model, exception_list=None, time_period=None):
+    for var in model.component_objects(Var):
```
nareshsusarla, line 23: I suppose, if you use `component_data_objects` instead of `component_objects`, you need not check if the var object is indexed and then it could simplify your code. Unless you need it for another reason

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,60 @@
...
+                if index is not None and time_period is not None:
+                    for i in index:
+                        if i in model.s_T and i not in time_period:
+                            is_in_time_period = False
+                            break
+                if not is_in_time_period:
+                    continue
+                index_var = var if index is None else var[index]
+                # unfix binary variables and unbound the continuous variables
+                if index_var.domain is Binary:
```
nareshsusarla, line 38: it may be better to use `is_binary()` method here

pareto/utilities/model_modifications.py
nareshsusarla: why both `unfix()` and `free()`, they are same methods?

pareto/utilities/bounding_functions.py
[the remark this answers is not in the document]
nareshsusarla: i  would think its best to remove it if the logic has changed

nareshsusarla (changes requested):

pareto/utilities/bounding_functions.py
```diff
@@ -0,0 +1,170 @@
...
+
+###############################################################################
+### Individual variable bound generation functions
+def Bound_v_F_Piped(model):
+    # initialization function for parameter
+    def p_F_Piped_UB_init(model, l, l_tilde):
+        # set default value to pipeline capacity plus maximum expansion capacity
+        return model.p_sigma_Pipeline[l, l_tilde] + max(
+            [value(model.p_delta_Pipeline[d]) for d in model.s_D]
+        )
```
nareshsusarla, line 58: **If the pipeline** connecting any two arcs **is bi-directional** and the model decides to install pipelines in both directions (this may be likely for trunk lines in new development areas), the choice of bound could be limiting the flow (at least conceptually)

nareshsusarla (changes requested):

pareto/utilities/model_modifications.py
tarnold17: Yep, it's a matter of style I guess.

pareto/utilities/model_modifications.py
tarnold17: I would also like an answer to @nareshsusarla's question here.

pareto/tests/test_utilities.py
```diff
@@ -0,0 +1,330 @@
...
+
+from pyomo.environ import value
+
+# Modules to test:
+from pareto.utilities.bounding_functions import VariableBounds
+from pareto.utilities.model_modifications import free_variables
+from pareto.utilities.model_modifications import deactivate_slacks
+
+# import pytest
+import pytest
```
tarnold17, line 37:
```suggestion
```
This import is unused, so these lines can be removed.

pareto/tests/test_utilities.py
tarnold17: @Tominapa your reasoning on 1 is fallacious. I was the second person to review this PR - you can't dismiss my question as unimportant simply because the first reviewer didn't mention it in their review. That said, while you could use a pytest fixture here, I don't think you need to - I was just wondering if there was a reason why you had chosen not to. You can learn more about fixtures [here](https://docs.pytest.org/en/6.2.x/fixture.html) and you can look at some of our other unit tests for examples of when/how to use them.

On 2 - as it is currently written, the function `build_strategic_model` allows you to build and solve the strategic model with whatever option you want for quality (as specified in the `config_dict` argument that the function takes). I understand why you need to solve the model in this function, so I suggest changing its name to `build_and_solve_strategic_model` to more accurately describe its behavior. In other parts of our codebase, functions that say "build" in the name generally just build a model without solving. So since this function both builds and also solves the model, you can prevent some potential confusion for other developers by changing the name to reflect that. This is admittedly a somewhat minor thing, but it's a change that will help anyone who reads this code somewhere down the line.

pareto/utilities/bounding_functions.py
Tominapa: To my knowledge there is not a "pipe" index that I could use to resolve this problem. That's a PARETO-level limitation as I see it. It won't be solved by these utility functions. I'm marking it as resolved without further consideration.

pareto/utilities/model_modifications.py
Tominapa: Perhaps it would.

pareto/utilities/model_modifications.py
Tominapa: Perhaps this is so.

pareto/utilities/model_modifications.py
Tominapa: The documentation for both is the same. In this sense, I see no harm in covering my bases. Perhaps a better question is this: Why does Pyomo have two identical methods?

pareto/utilities/bounding_functions.py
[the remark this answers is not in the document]
Tominapa: I might need it later if we are going to incorporate more variables, or change the logic for quality variables. This code is by no means complete at this stage.

pareto/tests/test_utilities.py
Tominapa: 1. Thanks for the pointer to the docs. I figured out why you are asking for the change based on the context there. It should be the same with the change, benefiting from any marginal test time reductions.
2. I see your point, but the key from my point of view is that I do not _want_ to solve the model at this stage. If I could construct the quality model without a solve call, I would. This would require a change within PARETO itself, which I would rather avoid - at least for the moment. I would also suggest that this function contains neither build nor solve calls directly. A more accurate name, by this logic, would simply be "fetch_strategic_model()"

Allow me some time to complete these changes. I have other priorities competing today.

tarnold17: @Tominapa:

1) to be clear, I don't think you _need_ to restructure this code to use a fixture. Feel free to do so if you wish, but I don't think it's strictly necessary.

2) Your function `build_strategic_model()` calls `create_model()` and then `solve_model()` - these are the functions that build the strategic model and then solve it, respectively, so I'm not sure how you can argue that `build_strategic_model()` "contains neither build nor solve calls directly". I understand that the fact that you need to call `solve()` to get the quality model is an artifact of how the strategic model is designed. I'm just saying that it's easier to read code where function names describe as closely as possible the actions of the function (even if some of those actions are side effects). To be clear, this is a relatively minor thing, and I'm not going to block you from merging this PR or anything if you don't want to change the name of the function, I just want to explain my opinion clearly.

pareto/utilities/bounding_functions.py
nareshsusarla: No, that is not correct, you need to adjust the value of the bound otherwise the value will be incorrect. Please review the two constraints in the model `PipelineCapacityExpansion` and `PipelineCapacity`. Although the index for pipeline is not modeled, but we it recognizes multiple pipelines through origin and destination and so, the capacity gets adjusted based on how many pipelines are there between 2 given nodes. An easiest way to do this, may be, is to add a factor of 2 to max() term in your method, i.e., 2 * max(..). A better way would be to identify the bi-directional pipelines and only apply the modification to them, your choice.

nareshsusarla (changes requested):

pareto/tests/test_utilities.py
Tominapa: Removed

pareto/tests/test_utilities.py
Tominapa: 1. Done.
2. Indeed, the build and solve calls are in those other functions. I don't want to give the impression that I am recreating those capabilities here. I'm just calling an existing function to get a model on which to run tests. The name is now "fetch_strategic_model()" to be compatible with your guidelines.

pareto/utilities/bounding_functions.py
melody-shellman: Good catch, Naresh!

melody-shellman (approved):
Good work here! I know  there will be a future iteration of bounding function, so I left a few notes in the description to help remind us what improvements we've already noted. Feel free to update with other items you think should be addressed for the next iteration.

tarnold17 (approved):

pareto/utilities/model_modifications.py
```diff
@@ -0,0 +1,59 @@
...
+                    index_var.unfix()
+                    index_var.free()
```
nareshsusarla, line 40:
```suggestion
                    index_var.unfix()
 ```

melody-shellman (approved):
Looks great! Good work !

nareshsusarla:
looks good

PR #273, opened 2023-10-05
Hydraulics Post_Process Infeasibility

melody-shellman:

- In release testing, it was identified that under config option hydraulics.post_process, the model was infeasible for the strategic toy case study
- Through testing, the team found that large pipeline distances resulted in pressure drops that exceeded the maximum limit, causing the infeasibility
- The current fix is reducing the pipeline distances in the case study, but this could be revisited for a more robust solution. 

### Legal Acknowledgement

contribution:

   employer.


nareshsusarla (commented):
I am guessing the objective value decreased because of the incentives from Beneficial reuse?

melody-shellman:
> I am guessing the objective value decreased because of the incentives from Beneficial reuse?

The CAPEX costs are calculated by  the pipeline distances, so the total CAPEX would have now decreased with the decreased distances. This is at least part of the difference. I did not check if there were other changes.

nareshsusarla (approved):
lgtm

tarnold17 (approved):

PR #258, opened 2023-06-28
154 update kpis for treatment

melody-shellman:
## Fixes/Addresses/Summary/Motivation:

- Adding Environmental KPIs
- Adding treatment KPIs to address #154 

## Changes proposed in this PR:
Identified and added environmental impact and environmental justice measures:
o	Air Pollutant Metrics. 
o	Trucking Activity
o	Additional measures not directly related to environmental impact

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my
contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.md file
   at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has
   rights to intellectual property that includes these contributions, I represent that I have
   received permission to make contributions and grant the required license on behalf of that
   employer.

PR #276, opened 2023-10-20
Support RKA and RKT arcs and beneficial reuse processing cost in strategic model

tarnold17:
Add support for RKA and RKT arcs and beneficial reuse processing cost to the strategic model.

New tabs in input files:
- "RKA"
- "RKT"
- "BeneficialReuseCost"

### Legal Acknowledgement

contribution:

   employer.


pareto/tests/test_strategic_model.py
melody-shellman, line 1: Why doesn't the test `test_run_toy_strategic_model` change with the addition of the costs? Particularly, the objective check: `assert pytest.approx(6122.5178, abs=1e-1) == pyo.value(m.v_Z)`?

melody-shellman (commented):
Looking good, just have a few questions!

pareto/case_studies/strategic_permian_demo.xlsx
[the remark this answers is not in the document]
tarnold17: @melody-shellman I think it is not desirable to add the headers here, for a couple reasons:

1) Avoid scope creep on this PR.
2) I based the branch for #281 off of the branch for this PR. So any further changes to input files would create a merge conflict in #281 that I'd need to resolve. That wouldn't be a huge problem, but I'd prefer to avoid it if possible.
3) At our software dev call this week, @lbianchi-lbl said he could look into a fix that would allow us to read tabs without headers using new Pandas without triggering an exception (the fix would probably involve a `try` / `except` block, as I understand). So it might not be necessary to update these tabs to include headers after all.

pareto/tests/test_strategic_model.py
tarnold17: @melody-shellman I took the values from the tab "BeneficialReuseCost" and added them to the values in the tab "BeneficialReuseCredit". So, the net credit of sending water to beneficial reuse (credit - cost) has not changed from before, and therefore we should not expect any of the solutions to change.

pareto/case_studies/strategic_permian_demo.xlsx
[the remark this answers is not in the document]
melody-shellman: oh great! I think best solution is what you mentioned here in 3.

pareto/tests/test_strategic_model.py
melody-shellman: got it! I figured that was the case, but just wanted to confirm.

melody-shellman (approved):
Looks good! Will need to update input sheets for all current PRs. Mostly relevant for #258.

ElmiraShamlou (approved):
LGTM

PR #282, opened 2023-11-22
Add files for seismicity risk metric calculation

HuidaeCho:
## Changes proposed in this PR:
This PR adds new files for seismicity risk metric calculation:
- `pareto/case_studies/strategic_toy_case_risk_study.xlsx`: Toy case study with Laura Capper's data
- `pareto/examples/strategic_model_risk_demo/SWD Risk Factor Logic.xlsx`: Laura Capper's original Excel with her logic
- `pareto/examples/strategic_model_risk_demo/strategic_model_risk_demo.ipynb`: An example run with a new function for seismicity risk metric calculation

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my
contribution:

1. I agree my contributions are submitted under the license terms described in the LICENSE.md file
   at the top level of this directory.
2. I represent I am authorized to make the contributions and grant the license. If my employer has
   rights to intellectual property that includes these contributions, I represent that I have
   received permission to make contributions and grant the required license on behalf of that
   employer.

PR #290, opened 2023-12-20
Add subsurface risk calculation to strategic_produced_water_optimization.py

HuidaeCho:
This PR adds subsurface risk metric calculation to
* `pareto/strategic_water_management/strategic_produced_water_optimization.py`

and add a case study input
* `pareto/case_studies/strategic_toy_case_risk_study.xlsx`

The subsurface risk tabs are populated for the following case study input file:
* `pareto/case_studies/workshop_baseline_all_data.xlsx`

### Legal Acknowledgement

contribution:

   employer.


ksbeattie:
This replaces #282

tarnold17 (changes requested):
This development looks good so far! I see a few changes which should be applied:

- Some of the tests in `pareto/tests/test_strategic_model.py` are currently failing because an additional configuration option has been added to the strategic model. This is a simple fix - the lines in the test file which have `assert len(m.config) == 8` just need to be changed to ``assert len(m.config) == 9`.
- At least one test should be added to `pareto/tests/test_strategic_model.py` to test the newly added SeismicityRisk option. You can look at other tests in this file to get an idea of what this test can look like.
- Black needs to be run on the added code (the "Run Black" check is currently failing).

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -6920,6 +6940,73 @@ def infrastructure_timing(model):
...
+            if factor in ("orphan", "inactive") and deep[site]:
+                site_risk_factor = distance_risk_factors[factor]
+            else:
+                site_risk_factor = min(prox[factor][site], distance_risk_factors[factor])
+            norm_factor = f"norm_{factor}"
+            weighted_site_risk_factor += site_risk_factor * distance_risk_factors[norm_factor] * severity_risk_factors[norm_factor]
+        inverse_weighted_site_risk_factor = weighted_site_risk_factor / max_risk_factor_for_lowest_risk
+        risk_metrics[site] = 1 - inverse_weighted_site_risk_factor
+        # only consider these sites with True sites_included!
+        sites_included[site] = pressure_thresholds["min"] <= pressure[site] < pressure_thresholds["max"]
```
tarnold17, line 7004:
```suggestion
        sites_included[site] = pressure_thresholds["min"] <= pressure[site] <= pressure_thresholds["max"]
```
Is there a reason to use different conventions for the lower and upper threshold? I think it makes sense to use the same convention for both.

HuidaeCho: That's based on Laura's logic in her Excel.
* Is average pressure around well <.5 psi/ft?  (0=no, 1= yes) (for nearby similar - deep or shallow -  wells)
* Is average pressure around well >=.7  psi/ft?  (0=no, 1= yes) (for nearby similar - deep or shallow -  wells)

If any of the above is 1, excluded. That's .5 psi/ft (min) <= pressure < .7 psi/ft means included.

tarnold17: @HuidaeCho I understand the logic, but why use `<=` when checking the lower threshold but `<` when checking the upper threshold? This is a minor thing, since it only has an effect if the pressure is exactly 0.5 or 0.7, but I think it would make more sense to be consistent when checking both thresholds. I would choose to use `<=` for both,  but using `<` for both would be fine as well.

pareto/case_studies/strategic_toy_case_risk_study.xlsx
tarnold17, line 1: I don't think we need to commit this file to the repository.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
```diff
@@ -0,0 +1,531 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "c6e5b284-d42e-4346-b353-11a4115adb00",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "min_cost_obj = model.v_Z.value\n",
+    "min_cost_risk = model.v_Z_SubsurfaceRisk.value\n",
+    "\n",
```
JavalVyas2000, line 260: With the changes of Pyomofication of the subsurface_risk_metric, one might have to change the min_cost_risk to be equal to `model.v_Z_SubsurfaceRisk.value + model.subsurface.y_dist.value` as there is a change in the objective function for the logic to work.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
```diff
@@ -0,0 +1,531 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "880b8561-b5f9-4470-acc0-fece09e40a03",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "min_risk_obj = model.v_Z_SubsurfaceRisk.value\n",
```
JavalVyas2000, line 313: `model.v_Z_SubsurfaceRisk.value + model.subsurface.y_dist`

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
```diff
@@ -0,0 +1,531 @@
...
+    "\n",
+    "# Create lists for results\n",
+    "fixed_risk_vals = list(range(0, 400, 20))  # Subsurface risk values to iterate over\n",
+    "cost_vals_fixed_risk = list()  # List to store corresponding cost values\n",
+    "\n",
+    "for r in fixed_risk_vals:\n",
+    "    print(\"*\" * 60)\n",
+    "    print(f\"Solving minimum cost model constraining subsurface risk to {r}\")\n",
+    "    print(\"*\" * 60)\n",
+    "    model.v_Z_SubsurfaceRisk.fix(r)  # Fix the subsurface risk to the specified value\n",
```
JavalVyas2000, line 352: Would suggest adding a constraint instead of this, as now model.v_Z_SubsurfaceRisk also includes the binary, the proposed constraint should be 

```
@model.Constraint()
def fix_risk(model):
      return model.v_Z_SubsurfaceRisk +sum(
                    model.subsurface.y_dist[i, j]
                    for i in model.subsurface.SWDSites
                    for j in prox
                )==r
```

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
```diff
@@ -0,0 +1,531 @@
...
+    "\n",
+    "min_cost_for_zero_risk = cost_vals_fixed_risk[0]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "326a34c6-1982-44c0-8b02-22ab1cfa7645",
+   "metadata": {},
+   "source": [
+    "We can plot the results as a Pareto curve (note that the concept of Pareto optimality inspired the name for Project PARETO):"
```
JavalVyas2000, line 376: Is this note necessary? As it might lead to someone believe at the the first glance that Project PARETO lead to the Pareto optimality.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
```diff
@@ -0,0 +1,531 @@
...
+    "\n",
+    "    assert results_min_risk.solver.termination_condition == TerminationCondition.optimal\n",
+    "    assert results_min_risk.solver.status == SolverStatus.ok\n",
+    "\n",
+    "    with nostdout():\n",
+    "        feasibility_status = is_feasible(model)\n",
+    "\n",
+    "    assert feasibility_status\n",
+    "\n",
+    "    risk_vals_fixed_cost.append(model.v_Z_SubsurfaceRisk.value)\n",
```
JavalVyas2000, line 442: `model.v_Z_SubsurfaceRisk.value - model.subsurface.y_dist.value`

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -30,6 +31,8 @@
     is_feasible,
     nostdout,
 )
+import pyomo.environ as pyo
+from idaes.core.util.model_statistics import degrees_of_freedom
```
JavalVyas2000, line 35: I do not think we need this import.

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -160,28 +164,30 @@
...
  infrastructure_timing: [InfrastructureTiming.false, InfrastructureTiming.true]
+ subsurface_risk: [SubsurfaceRisk.false, SubsurfaceRisk.exclude_risky_wells, SubsurfaceRisk.calculate_risk_metrics]
  """
 
 strategic_model = create_model(
     df_sets,
     df_parameters,
     default={
-        "objective": Objectives.cost,
+        "objective": Objectives.subsurface_risk,
```
JavalVyas2000, line 182: This should run with `cost` objective and not `subsurface_risk`

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -160,28 +164,30 @@
...
         "pipeline_cost": PipelineCost.distance_based,
         "pipeline_capacity": PipelineCapacity.input,
         "hydraulics": Hydraulics.false,
         "node_capacity": True,
         "water_quality": WaterQuality.false,
         "removal_efficiency_method": RemovalEfficiencyMethod.concentration_based,
-        "infrastructure_timing": InfrastructureTiming.true,
+        "infrastructure_timing": InfrastructureTiming.false,
```
JavalVyas2000, line 189: Like wise here as well, this should run with the original run_strategic_model.py options.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -4220,8 +4297,11 @@ def PumpCostRule(b, l1, l2):
         the cost of pumps.
         """
         # In the co_optimize method the objective should include the cost of pumps. So,
-        # delete the original objective to add a modified objective in this method.
-        del model.objective
+        # delete the original objectives to add a modified objective in this method.
+        del model.objective_Cost
```
JavalVyas2000, line 4301: If you deactivate then why do we delete the objective from the model? I am not sure how exactly the objective function is working. But I believe if we deactivate things then we should not delete objectives from the model

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -4353,8 +4442,16 @@ def PumpCostRule(b, l1, l2):
...
+        del model.objective_Reuse
+        if model.do_subsurface_risk_calcs:
+            del model.objective_SubsurfaceRisk
```
JavalVyas2000, line 4449: Similarly, here we delete the SubsurfaceRisk Objective not sure why.

JavalVyas2000 (commented):
I do not completely understand the way objectives are set and there might be a few changes in the MOO.ipynb file if we finalize on the pyomofied verison of calculating subsurface_risk_metric, but other than that thing PR looks good.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
tarnold17: @JavalVyas2000 I changed the subsurface risk objective function so that your suggestion is not necessary.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
tarnold17: I changed the subsurface risk objective function so that this suggestion is not necessary.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
tarnold17: I changed the subsurface risk objective function so that this suggestion is not necessary.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
tarnold17: I reworded the comment to try to clarify things.

pareto/examples/multiobjective_optimization/seismicity_vs_cost_MOO.ipynb
tarnold17: I changed the subsurface risk objective function so that this suggestion is not necessary.

pareto/strategic_water_management/run_strategic_model.py
tarnold17: Done.

pareto/strategic_water_management/run_strategic_model.py
tarnold17: Done.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Good point - instead of deleting, I changed the code to deactivate the original objective functions.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: Good point - instead of deleting, I changed the code to deactivate the original objective functions.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -4746,8 +4857,18 @@ def PumpCostRule(b, l1, l2):
         the cost of pumps.
         """
         # In the co_optimize method the objective should include the cost of pumps. So,
-        # delete the original objective to add a modified objective in this method.
-        del model.objective
+        # deactivate the original objectives to add a modified objective in this method.
+        model.objective_Cost.deactivate()
```
JavalVyas2000, line 4861: I believe we never activate the objective after deactivating them in 4711 - 4717. This part of the code may not be needed.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -7838,72 +8174,65 @@ def solve_model(model, options=None):
...
+        with open(f"{opt.options['solver']}.opt", "w") as f:
+            f.write(
+                f"$onecho > {opt.options['solver']}.opt\n optcr={gap}\n running_time={running_time} $offecho"
+            )
+
+    # Set maximum running time for solver
+    set_timeout(opt, timeout_s=running_time)
+
+    # Set solver gap
+    if opt.type in ("gurobi_direct", "gurobi"):
```
JavalVyas2000, line 8227: Should this not be an elif statement? I agree it would not make a difference in ideal case but can cause problems if there are multiple solvers parsed.

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -191,6 +192,7 @@
         "water_quality": WaterQuality.false,
         "removal_efficiency_method": RemovalEfficiencyMethod.concentration_based,
         "infrastructure_timing": InfrastructureTiming.true,
+        "subsurface_risk": SubsurfaceRisk.exclude_over_and_under_pressured_wells,
```
JavalVyas2000, line 195: I cannot comment on line 199, but should we add a doc string which would suggest what the options could be? I believe the following options are valid

- deactivate_slacks
- scale_model
- scaling_factor
- running_time
- gap
- solver
- only_subsurface_block

Not sure if anything else is a part of this options parameter.

JavalVyas2000 (commented):
This looks good to me apart from a couple of minor suggestions and questions.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -242,8 +252,21 @@ def get_strategic_model_unit_container():
...
+    "subsurface_risk",
+    ConfigValue(
+        default=SubsurfaceRisk.false,
+        domain=In(SubsurfaceRisk),
+        description="Subsurface risk",
+        doc="""Selection to include subsurface risk.
+        ***default*** - SubsurfaceRisk.false
+        **Valid Values:** - {
+        **SubsurfaceRisk.false** - Exclude subsurface risk from model (unless the subsurface risk objective function is selected),
+        **SubsurfaceRisk.exclude_over_and_under_pressured_wells** - Calculate subsurface risk metrics and disallow disposal to overpressured and underpressured wells,
```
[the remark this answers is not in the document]
tarnold17: @JavalVyas2000 the name used to be `exclude_risky_wells`. I changed it to `exclude_over_and_under_pressured_wells` as that is more descriptive.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -2306,75 +2320,104 @@ def init_arc_param(arctype):
...
-                )
-                + model.v_C_Slack
-                - model.v_R_TotalStorage
-                - model.v_R_TotalBeneficialReuse
-            )
+    ####### Maximum reuse objective #######
+    model.v_Z_Reuse = Var(
+        within=Reals,
+        units=pyunits.dimensionless,
+        doc="Objective function variable - maximize reuse [dimensionless]",
```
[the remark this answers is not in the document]
tarnold17: @JavalVyas2000 yes, this variable is dimensionless, as it is a ratio of the total amount of water reused to the total amount of flowback/produced water.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: @JavalVyas2000 the lines you are referring to are in different sections of an `if`/`elif` block - lines 4711-4717 corresponds to the config option for `Hydraulics.co_optimize` and 4861-4866 is in the section for `Hydraulics.co_optimize_linearized`. So it is necessary in both places.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: @JavalVyas2000 I don't think I am understanding your question here - can you clarify? This is an `if`/`elif` block that has a section for Gurobi and a section for CBC.

JavalVyas2000: I am saying it should be 
```
if CPLEX:
    Block of code
elif gurobi:
    Block of code
elif cbc:
    Block of code
else:
    Objective not supported. 
```

Here it is as follows:
```
if CPLEX:
    Block of code
if gurobi:
    Block of code
elif cbc:
    Block of code
else:
    Objective not supported. 
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -242,8 +252,21 @@ def get_strategic_model_unit_container():
...
+    ConfigValue(
+        default=SubsurfaceRisk.false,
+        domain=In(SubsurfaceRisk),
+        description="Subsurface risk",
+        doc="""Selection to include subsurface risk.
+        ***default*** - SubsurfaceRisk.false
+        **Valid Values:** - {
+        **SubsurfaceRisk.false** - Exclude subsurface risk from model (unless the subsurface risk objective function is selected),
+        **SubsurfaceRisk.exclude_over_and_under_pressured_wells** - Calculate subsurface risk metrics and disallow disposal to overpressured and underpressured wells,
+        **SubsurfaceRisk.calculate_risk_metrics** - Calcuulate subsurface risk metrics for the user to view, but don't change the optimization model
```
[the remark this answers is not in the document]
tarnold17: Addressed.

pareto/strategic_water_management/run_strategic_model.py
tarnold17: @JavalVyas2000 I agree it's a good suggestion to add documentation for the options for `solve_model()`. I have added a docstring to the function definition for `solve_model()` in `strategic_produced_water_optimization.py`.

pareto/strategic_water_management/strategic_produced_water_optimization.py
tarnold17: I see. It would be fine with me if you want to try to reformat the code in that manner, and confirm that it works with CPLEX (I don't have CPLEX installed, so I wouldn't be able to test that part easily).

pareto/strategic_water_management/strategic_produced_water_optimization.py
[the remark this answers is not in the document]
JavalVyas2000: I am still not convinced on this, I agree it is the ratio in the first part of the objective but not in the second part of the objective where the costs are added. Only way this can be dimensionless is if the factor 1/38446652 has the unit 1/kUSD. Also do you believe for consistency, we shall add units to other objective functions as well?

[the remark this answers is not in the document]
JavalVyas2000: I mistakenly took the deleted code to be the one that was existing. This comments are taken care in the current version of the code. Hence, resolved.

pareto/strategic_water_management/strategic_produced_water_optimization.py
JavalVyas2000: Based on the offline discussion, this should not be a problem going forward.

pareto/strategic_water_management/run_strategic_model.py
JavalVyas2000: Addressed in the solve_model function.

JavalVyas2000 (approved):
Looks good to me.

tarnold17 (approved):
Looks good!

Issue #90, opened 2021-11-22
Error Handling for Pipeline Config Options and Water Quality model configuration


melody-shellman:
### Description
The new config options for pipelines (pipeline_capacity, pipeline_cost) now require different tabs in the input file. We need a module to handle errors based on these input config options. At the moment, all tabs are required in order for the the code to run. The tests will also need to be updated. 

### Requirements
if pipeline_capacity = calculated, then the input file requires the tabs "Hydraulics", "PipelineDiameterValues"
if pipeline_capacity = input, then the input file requires the tab "PipelineCapacityIncrements"
if pipeline_cost = distance_based, then the input file requires the tabs "PipelineCapexDistanceBased", "PipelineExpansionDistance","PipelineDiameterValues"
if pipeline_cost = capacity_based, then the input file requires the tab  "PipelineCapexCapacityBased",

melody-shellman:
Note: good to also correct any remaining issues from #235

PR #291, opened 2023-12-21
Input Data Error Handling and Infeasibility Detection

melody-shellman:

There are three main goals of this PR:

- [x] Update get_data() to not require set_list and parameter_list. When given, behavior is same as before, when not given, get_data() reads all valid input tabs.
- [x] Raise warnings and/or errors when there are missing input tabs. (addresses issue #90 )
- [x] Raise warnings for infeasibility detections. 
- [x] Add test for build with missing input data

## Logic for required and non-required data:
**Minimum Required Data** - already implemented
- Required Sets: "ProductionPads","CompletionsPads", "PipelineDiameters",
- Required Parameters: "Units", "CompletionsDemand","FlowbackRates","InitialPipelineCapacity","PipelineOperationalCost",
- Required Parameter groups. Must include 1 or more within each list: ("PNA", "PCT", "PKT",),("CNA", "CCA", "CST", "CCT", "CKT"),("NCA", "FCA", "RCA", "SCA", "FCT",). The idea is that for each required "source" there is at least one type of  arc to remove water and for each type of  "sink" there is at least one type of arc to supply water. 

**Optional - can leave empty** - Done
- Optional: SWDSites. If SWDSites Set is given, then "InjectionCapacities", "InitialDisposalCapacity", "DisposalOperationalCost" are needed
- Optional: TreatmentSites. If Treatment sites are given, then  "TreatmentCapacities", "TreatmentTechnologies", "InitialTreatmentCapacity", "TreatmentOperationalCost",one of: ("RNA","RSA","RKA","RST","RKT"), "DesalinationTechnologies","DesalinationSites","TreatmentEfficiency" are required
- Optional: ReuseOptions. IF  "ReuseOptions", then one of( "ROA", "SOA", "NOA", "ROT", "SOT"),"BeneficialReuseCost","BeneficialReuseCredit","ReuseMinimum","ReuseCapacity","ReuseOperationalCost"
- Optional: Trucking. If any trucking arcs, then  "TruckingTime", "TruckingHourlyCost" are required. 
- Optional: StorageSites. "StorageCapacities", one of ("SOA", "SOT","SNA","SCA"), one of ("RSA"), "InitialStorageCapacity",
- Optional: Freshwater.  IF "FreshwaterSources", then one of("FCA", "FCT"), "FreshwaterSourcingAvailability", "FreshSourcingCost"
- Optional: "NNA","NKA","NRA","NSA","NetworkNodes". Never required, 
- Optional: CAPEX Expansion. IF ("DisposalCapacityIncrements", "StorageCapacityIncrements", ""TreatmentCapacityIncrements", or"PipelineCapacityIncrements"), then respective ExpansionCost data is required.  

**Optional - populate with default values** - Done
- "CompletionsPadOutsideSystem" (all inside system)
-  "Economics", (default values can be current data)
-  "DisposalOperatingCapacity", (all 100%)
- "PadOffloadingCapacity", (infinite)
- "CompletionsPadStorage", (0)

**Data Dependent on Configuration Options** - TODO
-  "objective": no specific data needed
- "pipeline_cost": PipelineCost.distance_based,

> if pipeline_cost = distance_based, then the input file requires the tabs "PipelineCapexDistanceBased", "PipelineExpansionDistance","PipelineDiameterValues"

> if pipeline_cost = capacity_based, then the input file requires the tab "PipelineCapexCapacityBased",
>  "pipeline_capacity": 

>  if pipeline_capacity = calculated, then the input file requires the tabs "Hydraulics", "PipelineDiameterValues"
> if pipeline_capacity = input, then the input file requires the tab "PipelineCapacityIncrements"

- "hydraulics":  "Hydraulics","Elevation","WellPressure","InitialPipelineDiameters","PipelineDiameterValues"
- "node_capacity": "NodeCapacities",
- "water_quality": "PadWaterQuality","StorageInitialWaterQuality","PadStorageInitialWaterQuality","RemovalEfficiency", #todo: may need to move the parameter read to be in water quality

-  "removal_efficiency_method": no specific data needed
-"infrastructure_timing":"TreatmentExpansionLeadTime","DisposalExpansionLeadTime","StorageExpansionLeadTime","pipeline_cost","PipelineExpansionLeadTime_Dist","PipelineExpansionLeadTime_Capac",

### Legal Acknowledgement

contribution:

4. I represent I am authorized to make the contributions and grant the license. If my employer has
   employer.


pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
 )
 
 
+# TODO: move this to utility
+class MissingDataError(Exception):
+    def __init__(self, message):
+        super().__init__(message)
+
+
+# TODO: Move this function to a utility
```
tarnold17, line 227: I agree with moving this to a utility. However, why define a function to return a constant list? Why not just assign the list to a variable instead?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+            )
+            + str(_missing_params)
+        )
+
+    return error_message
+
+
+# TODO: Move this function to a utility
+def check_required_data(df_sets, df_parameters, config):
+    # Create a  to hold all missing data that causes an error
```
tarnold17, line 375:
```suggestion
    # Create a list to hold all missing data that causes an error
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+    df_sets,
+    df_parameters,
+    config_argument,
+    config_required_sets,  # {config option 1: [set1, set2], config option 2: []}
+    config_required_params,
+):
+    # Error message
+    error_message = []
+
+    # Data error is raised if
```
tarnold17, line 344: This comment appears to be unfinished.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+
+    required_sets = set(config_required_sets[config_argument])
+    _missing_sets = required_sets - _df_sets_set
+
+    if len(_missing_sets) > 0:
+        error_message.append(
+            "The config option {} has been selected. The following sets are required for this option and are missing:".format(
+                config_argument
+            )
+            + str(_missing_sets)
```
tarnold17, line 356:
```suggestion
            f"The config option {config_argument} has been selected. The following sets are required for this option and are missing:"
            + str(_missing_sets)
```
A minor thing, but I recommend using an [f-string](https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/) instead of the `.format()` syntax. f-strings are a lot more readable, in my opinion.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+        if len(_missing_sets) > 0:
+            for s in _missing_sets:
+                df_sets[s] = {}
+            warning_message = "Warning: {} are given, but other {} Set data is missing. The following set has been assumed to be empty:".format(
+                optional_tab_name, optional_tab_name
+            ) + str(
+                _missing_sets
+            )
+            warnings.warn(warning_message)
+        if len(requires_at_least_one) > 0:
```
tarnold17, line 310:
```suggestion
```
This `if` statement is unnecessary.

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+                    )
+                    warnings.warn(warning_message)
+
+    # If there is specific parameter data for optional tab, the optional tab name is also required
+    if (
+        len(required_parameters_with_option & _df_parameters_set) > 0
+    ) and optional_tab_name not in df_sets:
+        raise MissingDataError(
+            'Essential data is incomplete. Parameters for {} are given, but the "{}" Set is not given. Please add and complete the following tab(s): {}'.format(
+                optional_tab_name, optional_tab_name, optional_tab_name
```
tarnold17, line 327: I'm not sure if this is correct - why is `optimal_tab_name` being printed at all three of the specified spots in this formatted string?

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+        PP_requires_at_least_one,
+    )
+
+    # SWDs
+    SWD_set_name = "SWDSites"
+    SWD_required_sets = {}
+    SWD_required_parameters = {"InitialDisposalCapacity", "DisposalOperationalCost"}
+    SWD_requires_at_least_one = [["NKA", "RKA", "PKT", "RKT"]]
+    (df_sets, df_parameters) = _check_optional_data(
+        df_sets, df_parameters, SWD_set_name, SWD_required_sets, SWD_required_parameters
```
tarnold17, line 536:
```suggestion
        df_sets, df_parameters, SWD_set_name, SWD_required_sets, SWD_required_parameters, SWD_requires_at_least_one
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+        "SOA",
+        "ROA",
+    ]
+
+
+# TODO: Move this function to a utility
+def _check_optional_data(
+    df_sets,
+    df_parameters,
+    optional_tab_name,
```
tarnold17, line 281: Is this `optional_tab_name` argument meant to always correspond to a set instead of a parameter? If so, I think the argument name should reflect that. Also, I think the word "optional" in the name is confusing here, because the tab name that is passed in this argument is often something which is in fact not optional (e.g., "CompletionsPads").

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -215,6 +218,489 @@ def get_strategic_model_unit_container():
...
+        missing_trucking_data = set(trucking_parameters) - set(df_parameters)
+        if len(missing_trucking_data) > 0:
+            # Add empty dictionary to df_parameters
+            for s in missing_trucking_data:
+                df_parameters[s] = {}
+            warning_message = "Warning: {} are given, but some trucking parameters are missing. The following missing parameters have been set to default values:".format(
+                "Trucking Arcs"
+            ) + str(
+                missing_trucking_data
+            )
```
tarnold17, line 680:
```suggestion
            warning_message = "Warning: Trucking arcs are given, but some trucking parameters are missing. The following missing parameters have been set to default values:" + str(
                missing_trucking_data
            )
```

pareto/utilities/get_data.py
```diff
@@ -53,25 +163,34 @@ def _read_data(_fname, _set_list, _parameter_list):
 
     _df_parameters = pd.read_excel(
```
tarnold17, line 164: Since we are reading in all tabs with the call to `pd.read_excel()`, it seems unnecessary to call this function twice. I recommend reformatting the code to call `pd.read_excel()` only once, and then extract the sets and parameters.

pareto/utilities/get_data.py
```diff
@@ -53,25 +163,34 @@ def _read_data(_fname, _set_list, _parameter_list):
...
+        if key in valid_parameter_tab_names
+    }
+
+    # Cleaning Parameters.
     # A parameter can be defined in column format or table format.
     # Detect if columns which will be used to reshape the dataframe by defining
     # what columns are Sets or generic words
-    # If _set_list is empty, it is assummed that a parameter is column format is being read.
-    # and _set_list is created based on the DataFrame column names, except for the last name,
+    # If valid_set_tab_names is empty, it is assumed that a parameter is column format is being read.
```
tarnold17, line 185:
```suggestion
    # If valid_set_tab_names is empty, it is assumed that a parameter in column format is being read.
```

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: I like the function because it feels cleaner when calling from different files. The intention is for this list to be defined only once throughout PARETO and accessed in other files with the get_..() function

FYI  commit is coming, but quite a lot is changing since this versions, so replying to these comments

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: Good note. I updated the name to include "set". I do believe these will all be optional. As of now, we require no Sets in PARETO, we do however require a number of "one of " sets. For example, one of ["ProductionPads", "CompletionsPads", "FreshwaterSources"] (water sources). This can be modified pretty easily though if we want to change what is required.

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: Applied locally, good note!

pareto/strategic_water_management/strategic_produced_water_optimization.py
melody-shellman: I updated it slightly. This warning displays if the optional_set_name is missing from the input and required data dependent on it is included (e.g. "SWDSites" is missing, but "DisposalOperationalCost" is included). The warning will look like:

'Essential data is incomplete. Parameter data for {SWDSites} is given, but the "{SWDSites}" Set is missing. Please add and complete the following tab(s): {SWDSites}, or remove the following Parameters:{DisposalOperationalCost}'

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+    strategic_model.p_beta_TotalPW = Param(
+        strategic_model.s_T,
+        units=strategic_model.model_units["volume"],
+        doc="Combined water supply forecast (flowback & production) over the planning horizon [volume]",
+        mutable=True,  # Mutable Param - can be changed in sensitivity analysis without rebuilding the entire model
+    )
+
+    # # Get the maximum capacity of sink nodes for each time period
+    # # Sink nodes: CompletionsDemand, (InitialDisposalCapacity + max(disposalCapacityIncrements)) * DisposalOperatingCapacity,
+    # # StorageEvaporation, ReuseCapacity, InitialStorageCapacity + Storage CapacityIncrements,
```
tarnold17, line 605:
```suggestion
    # Get the maximum capacity of sink nodes for each time period
    # Sink nodes: CompletionsDemand, (InitialDisposalCapacity + max(disposalCapacityIncrements)) * DisposalOperatingCapacity,
    # StorageEvaporation, ReuseCapacity, InitialStorageCapacity + Storage CapacityIncrements,
```

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+                        (
+                            strategic_model.p_sigma_Treatment[r, wt].value
+                            + max(
+                                [
+                                    strategic_model.p_delta_Treatment[wt, j].value
+                                    for j in strategic_model.s_J
+                                ]
+                            )
+                        )
+                        * (1 - strategic_model.p_epsilon_Treatment[r, wt].value)
```
tarnold17, line 676:
```suggestion
```
I don't think you need this line. By including it, you are essentially discounting the treatment residual from the total PW capacity of the network.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+                        for wt in strategic_model.s_WT
+                    ]
+                )
+                for r in strategic_model.s_R
+            )  # Treatment residual: for each treatment site - treatment technology that yields the
+            # maximum value for: (initial treatment + max treatment expansion) * treatment efficiency
```
tarnold17, line 682:
```suggestion
            # maximum value for: (initial treatment + max treatment expansion)
```

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+        )
+        # Note: if completions pad is outside system, demand is not required to be met
+        strategic_model.p_gamma_TimePeriodDemand[t] = sum(
+            strategic_model.p_gamma_Completions[cp, t]
+            * (1 - strategic_model.p_chi_OutsideCompletionsPad[cp])
```
tarnold17, line 687:
```suggestion
            if strategic_model.p_chi_OutsideCompletionsPad[cp] == 0 else 0
```
I think this is a more straightforward way to write this logic.

tarnold17 (commented):
I have a number of suggestions and questions. On the whole this is really nice work; these features should make PARETO easier to use, especially for newer users.

pareto/models_extra/CM_module/case_studies/CM_large_permian.xlsx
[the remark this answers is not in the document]
melody-shellman: Good question -
"FreshwaterSources" changed to "ExternalWaterSources"
"FreshwaterSourcingAvailability" changed to "ExtWaterSourcingAvailability"
"FreshSourcingCost" changed to "ExternalSourcingCost"
Note: these changes are needed in both tab names and also subsequent tabs that reference the tab names (e.g. FCA)

pareto/models_extra/CM_module/case_studies/CM_small_permian.xlsx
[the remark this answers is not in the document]
melody-shellman: Good question -
"FreshwaterSources" changed to "ExternalWaterSources"
"FreshwaterSourcingAvailability" changed to "ExtWaterSourcingAvailability"
"FreshSourcingCost" changed to "ExternalSourcingCost"
Note: these changes are needed in both tab names and also subsequent tabs that reference the tab names (e.g. FCA)

pareto/tests/strategic_toy_case_study_data_error.xlsx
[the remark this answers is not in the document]
melody-shellman: These are new testing files (note the "error" label). There are comments in test_utilities.py . In this case, this file is is missing: 'Units', All of tabs ['ProductionPads', 'CompletionsPads', 'ExternalWaterSources'], and 'PipelineCapexCapacityBased' which is required for PipelineCost.capacity_based:{} which will cause an error in the data check.

pareto/utilities/get_data.py
```diff
@@ -23,19 +23,182 @@
...
+            and name != "Overview"
+            and name != "Schematic"
+        ):
+            unused_tab_list.append(name)
+
+    if len(unused_tab_list) > 0:
+        warning_message = (
+            f"Invalid PARETO input has been provided and has NOT been parsed. Check that input tab names match valid PARETO input. If you'd like to read custom tabs, please include as input to get_data(). The following tabs are NOT parsed: "
+            + str(unused_tab_list)
+        )
```
JavalVyas2000, line 187: I believe the same warning message appears after the results saying that the result excel headers are invalid pareto inputs. Not sure where this function is being called again.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+                df_parameters[param] = {}
+            for s in _missing_sets:
+                df_sets[s] = {}
+            _missing_tabs = _missing_sets
+            _missing_tabs.update(_missing_parameters)
+            warning_message = (
+                f"{optional_set_name} are given, but {optional_set_name} data is missing. The following tabs have been set to default values:"
+                + str(_missing_tabs)
+            )
+            warnings.warn(warning_message)
```
JavalVyas2000, line 779: I believe adding a stacklevel=3 would be better to avoid warnings.warn(warning_message) being printed.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+        # For each group in requires_at_least_one, check that at least one parameter exists in the input
+        for requires_list in requires_at_least_one:
+            _included_params = set(requires_list) & _df_parameters_set
+            if len(_included_params) < 1:
+                warning_message = (
+                    f"{optional_set_name} are given, but some piping and trucking arcs are missing. At least one of the following arcs are required, missing sets have been assumed to be empty:"
+                    + str(requires_list)
+                )
+                warnings.warn(warning_message)
```
JavalVyas2000, line 788: stacklevel=3 to avoid warnings.warn() being printed.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+        if param_tab not in df_parameters.keys():
+            df_parameters[param_tab] = {}
+            default_used.append(param_tab)
+    # Raise warning for default values
+    if len(default_used) > 0:
+        warning_message = (
+            f"The following parameters were missing and default values were substituted:"
+            + str(default_used)
+        )
+        warnings.warn(warning_message)
```
JavalVyas2000, line 582: stacklevel=3 to avoid warnings.warn() being printed.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+    return (df_sets, df_parameters)
+
+
+def model_infeasibility_detection(strategic_model):
+    # check_required_data() performs basic feasibility checks:
+    # - at least one source node in the model
+    # - at least on sink node in the model
+    # - each source node has an outgoing arc, each sink node has an incoming arc, and all other nodes have both an incoming and an outgoing arc.
+
```
JavalVyas2000, line 591: This is a suggestion, if possible plot the network graph using networkx. This will give a good visual representation of the model and make it possible for the user to check if there are any missing arcs without going through the excel sheet. I can help with this implementation. The following resource has a [tutorial](https://networkx.org/documentation/stable/tutorial.html) for the same.

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -220,4 +111,5 @@
 set_list = []
```
JavalVyas2000, line 111: I believe this should not be here as I do not see why the get_data() function is called at the end of the file again. Should it not be read before hand and the create a model and solve it or this is done on purpose?

pareto/utilities/get_data.py
```diff
@@ -195,6 +370,9 @@ def get_data(fname, set_list, parameter_list, sum_repeated_indexes=False):
       The header should start in row 2, and the index column should start in cell A3.
       Column format: Does not require a header. Each set should be placed in one column,
       starting from column A and row 3. Data should be provided in the last column.
+    - set_list and parameter_list are optional parameters. When they are not given, tabs with
+      valid PARETO labels are read. Otherwise, only the specified tabs in set_list and
+      parameter_list are read.
```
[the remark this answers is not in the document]
JavalVyas2000: Correct me if I am wrong, but if there a use of reading this custom tabs as create model does not have this parameters and variables in them and thus would not be used to create and solve the model.

pareto/strategic_water_management/run_strategic_model.py
melody-shellman: get_data() was developed as a flexible function to read in both PARETO input AND PARETO output. These lines at the end of this file demonstrate how get_data() can be used to read PARETO output, if desired. This requires passing in custom set_list and custom parameter_list. 
This example was not introduced in this PR. If it's something we'd like removed, let's plan for it in a separate PR!

pareto/tests/test_utilities.py
```diff
@@ -393,8 +290,173 @@ def test_fix_vars():
...
-    test_utilities_wout_quality()
-    test_utilities_w_post_quality()
-    test_utilities_w_discrete_quality()
-    test_fix_vars()
+def test_data_check():
+    # This emulates what the pyomo command-line tools does
+    # Check that MissingDataError is correctly raised
+    # This input sheet is missing: 'Units',
+    # All of tabs ['ProductionPads', 'CompletionsPads', 'ExternalWaterSources'],
+    # and 'PipelineCapexCapacityBased' which is required for PipelineCost.capacity_based:{}
```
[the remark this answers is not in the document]
melody-shellman: I phrased it this way because each of those groupings causes separate warnings/errors. It is confusing for a reader, though, so I like your rephrasing.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+# Raise errors and warnings for missing data and infeasibilities.
+def check_required_data(df_sets, df_parameters, config):
+    # Import config options within module to avoid circular module imports
+    from pareto.strategic_water_management.strategic_produced_water_optimization import (
+        WaterQuality,
+        PipelineCost,
+        Hydraulics,
+        PipelineCapacity,
+        InfrastructureTiming,
+    )
```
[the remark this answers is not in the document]
melody-shellman: Unfortunately these must be defined within the function, otherwise it results in circular imports (imports from strategic_produced_water..., but strategic also imports from this file). The import here allows imports so the files can be separate which I feel is cleaner.

pareto/utilities/get_data.py
```diff
@@ -23,19 +23,180 @@
...
     This methods uses Pandas methods to read from an Excel spreadsheet and output a data frame
     Two data frames are created, one that contains all the Sets: _df_sets, and another one that
     contains all the parameters in raw format: _df_parameters
     """
+    pareto_input_set_tab_names = get_valid_input_set_tab_names()
+    pareto_input_parameter_tab_names = get_valid_input_parameter_tab_names()
+
+    if _set_list is not None:
+        valid_set_tab_names = pareto_input_set_tab_names
+        valid_set_tab_names.extend(_set_list)
```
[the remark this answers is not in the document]
melody-shellman: Good note. While duplicated tab names do not necessarily cause issue with how it is currently implemented, I do think it makes sense to ensure the valid tab lists do not have duplicates.

pareto/strategic_water_management/run_strategic_model.py
JavalVyas2000: I believe this is the reason for the warnings that a user would get for the invalid pareto tabs. I am not sure if this should be read or not.

melody-shellman: Yes, this is the exact line that causes the warnings - I think the current version of main also does something similar? With the code you introduced, a warning is raised for unused tabs which would cause similar warning with reading the PARETO output.

pareto/utilities/get_data.py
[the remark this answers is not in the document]
melody-shellman: Good comment - here is a rephrased sentence that I will push: "Invalid PARETO input has been provided and has NOT been parsed. Check that the input tab names match valid PARETO input. If you'd like to read custom tabs (e.g., PARETO output files), please pass a list of the custom tab names to get_data(). The following tabs are NOT parsed: " Is this better?

pareto/strategic_water_management/run_strategic_model.py
JavalVyas2000: Yes. I had missed that there. I thought that was some other warning, but with a more comprehensive warning message figured that it is the reason. I am not sure if we should have this in main. Not exactly sure about what is it doing there? Probably @tarnold17 can answer this question.

pareto/utilities/get_data.py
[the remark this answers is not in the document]
JavalVyas2000: Yes this is good. Thanks.

pareto/models_extra/CM_module/cm_utils/data_parser.py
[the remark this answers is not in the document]
melody-shellman: Good one! I didn't check for a jupyter notebook. I'm now wondering if I should do the same for all of our jupyter notebooks.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+    df_sets,
+    df_parameters,
+    config_argument,  # e.g., "Hydraulics"
+    config_required_sets,  # {config option 1: [set1, set2], config option 2: []}
+    config_required_params,  # {config option 1: [param1], config option 2: [param2]}
+):
+    # Error message
+    error_message = []
+
+    # Data error is raised if
```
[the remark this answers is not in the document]
melody-shellman: Updated!

pareto/utilities/get_data.py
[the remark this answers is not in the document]
melody-shellman: @tarnold17  - good catch, this was how I initially implemented it, but I adjusted it after the first implementation. 
@JavalVyas2000 - The custom tabs are used in a few examples in the documentation (reading PARETO output files). It could also be used by developers who want to introduce a new parameter quickly for testing.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+        config.water_quality is WaterQuality.discrete
+        or config.water_quality is WaterQuality.post_process
+    ):
+        # Check if storage sites are given. If so, check for storage initial water quality.
+        (df_sets, df_parameters) = _check_optional_data(
+            df_sets,
+            df_parameters,
+            optional_set_name="StorageSites",
+            required_sets_with_option={},
+            required_parameters_with_option={"StorageInitialWaterQuality"},
```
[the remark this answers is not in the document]
melody-shellman: If storageSites is given, "StorageInitialWaterQuality" is required, HOWEVER, if it is left out, an empty dictionary is passed, and  a default value is applied. I can understand the confusion, but I couldn't quite find a word that better describes it. Any ideas?

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,838 @@
...
+        df_parameters,
+        optional_set_name="StorageSites",
+        required_sets_with_option={},
+        required_parameters_with_option={
+            "InitialStorageCapacity",
+            "InitialStorageLevel",
+            "StorageCost",
+            "StorageWithdrawalRevenue",
+        },
+        requires_at_least_one=[["SOA", "SOT", "SNA", "SCA"], ["RSA"]],
```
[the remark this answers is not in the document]
melody-shellman: You are correct. I will double check all of these. I think there's even one more for this.

pareto/utilities/process_data.py
melody-shellman: See previous comment.

pareto/utilities/process_data.py
melody-shellman: It depends on if we think of treatment centers as an endpoint for water OR a node that has some loss. 
Here I've treated it as a node that has some loss. Looking at feasibility, we then want to only include the residual as the end point for water not the full volume of treated water. 
I think we tend to view treatment sites more as a node with loss as water cannot end at the treatment site. 

Does you agree?

pareto/utilities/get_data.py
melody-shellman: Same comment as earlier, so resolving this one!

pareto/utilities/process_data.py
melody-shellman: I like this! I've created a new issue to track this suggestion.

pareto/strategic_water_management/run_strategic_model.py
melody-shellman: It's existed as long as I have been on the project. I think the run file used to be more of a tutorial than a true run file like it has evolved into now. Maybe it's not worth to have here since we have documentation and jupyter notebooks to demonstrate. I'm OK with removing- @tarnold17 , thoughts?

pareto/utilities/process_data.py
melody-shellman: Created issue #321 to be covered in a separate PR.

pareto/utilities/get_data.py
melody-shellman: Same comment as other discussion. Resolving this conversation!

pareto/utilities/process_data.py
melody-shellman: I like it!

pareto/utilities/get_data.py
```diff
@@ -195,6 +372,9 @@ def get_data(fname, set_list, parameter_list, sum_repeated_indexes=False):
       The header should start in row 2, and the index column should start in cell A3.
       Column format: Does not require a header. Each set should be placed in one column,
       starting from column A and row 3. Data should be provided in the last column.
```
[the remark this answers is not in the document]
melody-shellman: Created a new issue #322

pareto/strategic_water_management/run_strategic_model.py
melody-shellman: Per our conversation, removed!

pareto/utilities/get_data.py
melody-shellman: Looking at this again, the way it was designed is intentional. Sets and parameters are read in with different settings of `pd.read_excel` and are subsequently processed a bit differently.

JavalVyas2000 (commented):
Looks good to me. Apart from few minor suggestions.

pareto/examples/strategic_model_general_demo/strategic_model_general_demo.ipynb
```diff
@@ -104,135 +103,15 @@
...
+    "\n",
+    "PARETO's `get_data` utility function ([documentation](https://pareto.readthedocs.io/en/latest/utilities/Get_Data.html#get-data-get-data)) called below reads the data from the input file into the `df_sets` and `df_parameters` variables. The variables `df_sets` and `df_parameters` which are returned are dictionaries where the keys are the workbook tab names and the values are either dictionaries or Pandas dataframe objects containing the data read from the corresponding workbook tab. By default `get_data` will only read approved PARETO input tabs. Custom data tabs can optionally be passed to the function, Step 5 demonstrates an example of customizing `get_data` to read PARETO output files. "
    ]
```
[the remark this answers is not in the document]
melody-shellman: This is a good point. I've added a note in both the jupyter notebook and the get_data.rst file.

pareto/examples/strategic_model_general_demo/strategic_model_general_demo.ipynb
```diff
@@ -474,7 +353,7 @@
    "id": "4dab9b19",
    "metadata": {},
    "source": [
-    "The `get_data` function can also be used to load data from a previously generated PARETO results report back into Python:"
+    "The `get_data` function can also be used to load data from a previously generated PARETO results report back into Python. The `get_data` function is primarily designed to read PARETO input files. Because generated reports are not valid PARETO input, the desired tabs to be read must be explicitly passed to `get_data`:"
    ]
```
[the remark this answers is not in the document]
melody-shellman: I've added a note to the jupyter notebook for this!

pareto/strategic_water_management/strategic_produced_water_optimization.py
```diff
@@ -278,6 +285,11 @@ def create_model(df_sets, df_parameters, default={}):
     model.df_sets = df_sets
     model.df_parameters = df_parameters
 
+    # check that input data contains required data
+    model.df_sets, model.df_parameters = check_required_data(
+        model.df_sets, model.df_parameters, model.config
+    )
```
tarnold17, line 291:
```suggestion

    # check that input data contains required data
    model.df_sets, model.df_parameters = check_required_data(
        df_sets, df_parameters, model.config
    )
```

pareto/strategic_water_management/run_strategic_model.py
```diff
@@ -227,9 +115,3 @@
     output_units=OutputUnits.user_units,
     fname="strategic_optimization_results.xlsx",
 )
-
-# This shows how to read data from PARETO reports
-set_list = []
-parameter_list = ["v_F_Trucked", "v_C_Trucked"]
-fname = "strategic_optimization_results.xlsx"
-[sets_reports, parameters_report] = get_data(fname, set_list, parameter_list)
```
tarnold17, line 235: What is the reason for removing these lines?

pareto/utilities/get_data.py
```diff
@@ -54,25 +236,34 @@ def _read_data(_fname, _set_list, _parameter_list):
...
+
+    # Cleaning Parameters.
     # A parameter can be defined in column format or table format.
     # Detect if columns which will be used to reshape the dataframe by defining
     # what columns are Sets or generic words
-    # If _set_list is empty, it is assummed that a parameter is column format is being read.
-    # and _set_list is created based on the DataFrame column names, except for the last name,
+    # If valid_set_tab_names is empty, it is assumed that a parameter in column format is being read.
+    # and valid_set_tab_names is created based on the DataFrame column names, except for the last name,
     # which is used as the data column name.
```
tarnold17, line 260: The way the code is written, it is impossible for `valid_set_tab_names` to be empty. Can you please rewrite these comments to be a little clearer about what is going on here?

tarnold17 (commented):
I haven't reviewed everything yet but I here are the comments I have so far.

pareto/strategic_water_management/run_strategic_model.py
JavalVyas2000: The reason was this read the results which caused warnings as the results files do not have standard pareto input tab names.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,932 @@
...
+    capacity_feasibility_message = []
+    demand_feasibility_message = []
+    # Checks that for each time period, the volume of produced water is below the maximum capacity limit of sink nodes
+    # Get the total system produced water for each time period (PadRates, FLowbackRates)
+    strategic_model.p_beta_TotalPW = Param(
+        strategic_model.s_T,
+        units=strategic_model.model_units["volume"],
+        doc="Combined water supply forecast (flowback & production) over the planning horizon [volume]",
+        mutable=True,  # Mutable Param - can be changed in sensitivity analysis without rebuilding the entire model
+    )
```
tarnold17, line 695: Why is this implemented as a mutable parameter? I don't see any benefit to using a mutable `Param` over just using a Python variable. You could use a Pyomo expression to do the calculation - I think that would be sensible.

This comment applies to all of the `Params` which are defined in ``model_infeasibility_detection().

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,932 @@
...
+        doc="Total PW and external water available [volume]",
+        mutable=True,  # Mutable Param - can be changed in sensitivity analysis without rebuilding the entire model
+    )
+
+    for t in strategic_model.s_T:
+        strategic_model.p_beta_TotalPW[t] = sum(
+            strategic_model.p_beta_Production[p, t]
+            + strategic_model.p_beta_Flowback[p, t]
+            for p in strategic_model.s_P
+        )
```
tarnold17, line 727: E.g., this could be implemented as a Pyomo expression.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,932 @@
...
+                            )
+                        )
+                        for wt in strategic_model.s_WT
+                    ]
+                )
+                for r in strategic_model.s_R
+            )  # Treatment: for each treatment site, select treatment technology that yields the
+            # maximum value for (initial treatment + max treatment expansion)
+        )
+        # Note: if completions pad is outside system, demand is not required to be met
```
tarnold17, line 777:
```suggestion

        # Note: if completions pad is outside system, demand is not required to be met
```
Newline for code readability.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,932 @@
...
+        )
+
+        # If volume of produced water is greater than capacity, raise an infeasibility error
+        if (
+            strategic_model.p_beta_TotalPW[t].value
+            > strategic_model.p_sigma_MaxPWCapacity[t].value
+        ):
+            capacity_feasibility_message.append(
+                f"{t} ({round(strategic_model.p_beta_TotalPW[t].value)} total {pyunits.get_units(strategic_model.p_beta_TotalPW[t])}s PW)"
+            )
```
tarnold17, line 798: I think it would be a good idea to include the value of `p_sigma_MaxPWCapacity[t]` in the error message, so the user can easily see how bad the infeasibility is.

pareto/utilities/process_data.py
```diff
@@ -0,0 +1,932 @@
...
+
+        # If completions demand is greater than combined water in system and external water available, raise an infeasibility error
+        if (
+            strategic_model.p_gamma_TimePeriodDemand[t].value
+            > strategic_model.p_sigma_WaterAvailable[t].value
+        ):
+            demand_feasibility_message.append(
+                f"{t} ({round(strategic_model.p_gamma_TimePeriodDemand[t].value)} {pyunits.get_units(strategic_model.p_gamma_TimePeriodDemand)}s demand)"
+            )
```
tarnold17, line 807: I think it would be a good idea to include the value of `p_sigma_WaterAvailable[t]` in the error message, so the user can easily see how bad the infeasibility is.

tarnold17 (commented):
More comments and suggestions, mostly minor. I'm still not done reviewing, but I am done going through the densest part (`process_data.py`).

pareto/tests/strategic_toy_case_study_data_error.xlsx
tarnold17, line 1: The header cell in a lot of the tabs of this workbook have a #REF! error, e.g.: 
image

tarnold17 (changes requested):
More comments/suggestions.

JavalVyas2000:
@tarnold17 which is the jupyter notebook in which you are getting error in cell 6?

pareto/tests/strategic_toy_case_study_data_error.xlsx
melody-shellman: This is expected. This file was created for a test. This specific test checks that an error is correctly raised when mandatory data is missing. In this case, the "Units" tab was removed from the file (which should raise an error). The references are broken because they rely on the "Units" tab.

pareto/strategic_water_management/run_strategic_model.py
melody-shellman: Agree with Javal's comment. It was odd to have it in the run_strategic_file (mostly for developers). This example is in the documentation and in a jupyter notebook example.

pareto/utilities/process_data.py
melody-shellman: Done!

pareto/utilities/process_data.py
melody-shellman: Done!

pareto/utilities/get_data.py
melody-shellman: Good observation. Theses changes were originally to update comments to match the updates variables used in the function. Looking at it again, I think this section again, I think it may actually be problematic with the new set-up of get_data. I will remove and test. 
To give a short description of this section of code - it allowed users to pass in dataset with no sets and parameters in a column-only format. It would then assume the column names were valid sets. We don't have any datasets that follow this format and I wasn't even aware that this was a feature of get_data(). IT would also take some effort to make it work with the new get_data, so I'm opting to remove.

pareto/utilities/process_data.py
melody-shellman: Reasonable point - majority of our Params in PARETO are mutable, so I started here, but I see your point. These parameters are dependent on other params (e.g. defined by expressions), so they should not separately be mutable. Updated.

pareto/utilities/process_data.py
melody-shellman: Done!

pareto/utilities/get_data.py
```diff
@@ -23,19 +23,196 @@
...
+    unused_tab_list = []
+    df = pd.ExcelFile(_fname)
+    sheet_list = df.sheet_names
+    for name in sheet_list:
+        if (
+            name not in valid_set_tab_names
+            and name not in valid_parameter_tab_names
+            and name != "Overview"
+            and name != "Schematic"
+        ):
```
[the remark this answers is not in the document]
melody-shellman: Items in `_set_list` and `_parameter_list` will already be included in valid_set_tab_names! So a good suggestions, but happens to already be implemented :)

pareto/models_extra/CM_module/cm_utils/run_utils.py
```diff
@@ -32,16 +32,12 @@
...
-    set_list,
-    parameter_list,
-)
 from pareto.utilities.solvers import get_solver
 
 
 def load_data(fpath):
     # importing data
-    [df_sets, df_parameters] = get_data(fpath, set_list, parameter_list)
+    [df_sets, df_parameters] = get_data(fpath)
```
melody-shellman, line 40: need to update for CM model type

pareto/utilities/get_data.py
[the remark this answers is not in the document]
melody-shellman: I haven't addressed this yet - feels lower priority to me. Would you be able to add this?

melody-shellman:
> @tarnold17 I too ran into the error at cell 6 of screening_tool notebook. I am not sure why this having an error either.

This should be fixed now!

pareto/models_extra/CM_module/cm_utils/data_parser.py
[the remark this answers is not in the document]
tarnold17: It's working fine now - resolving.

pareto/utilities/get_data.py
```diff
@@ -22,20 +22,257 @@
...
+    "PadOffloadingCapacity",
+    "CompletionsPadStorage",
+    "ReuseOperationalCost",
+    "ExternalWaterQuality",
+    "InitialStorageCapacity",
+    "RemovalEfficiency",
+    "StorageCost",
+    "StorageWithdrawalRevenue",
+    "BeneficialReuseCost",
+    "BeneficialReuseCredit",
```
tarnold17, line 79:
```suggestion
```
Remove duplicate entries from list.

pareto/utilities/get_data.py
```diff
@@ -22,20 +22,257 @@
...
+    "InitialPipelineCapacity",
+    "InitialDisposalCapacity",
+    "InitialTreatmentCapacity",
+    "InitialStorageLevel",
+    "InitialStorageCapacity",
+    "RemovalEfficiency",
+    "StorageCost",
+    "StorageWithdrawalRevenue",
+    "BeneficialReuseCost",
+    "BeneficialReuseCredit",
```
tarnold17, line 133:
```suggestion
```
Remove duplicate entries from list.

pareto/utilities/process_data.py
tarnold17: This looks good to me with the latest commits. Resolving.

pareto/utilities/get_data.py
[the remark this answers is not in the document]
tarnold17: I agree that this is lower priority and doesn't need to be addressed immediately. I've opened issue #326 for us to refer back to later.

tarnold17 (approved):
Excellent work!

JavalVyas2000 (approved):
Great work Melody. This looks good to me.

PR #305, opened 2024-02-27
Add earthquake distance calculators

HuidaeCho:

This PR implements an earthquake distance calculator using the USGS and TexNet APIs for subsurface risk analysis.

- Add `calculate_earthquake_distances_usgs()` and `calculate_earthquake_distances_texnet()` functions
-

### Legal Acknowledgement

contribution:

   employer.


pareto/utilities/earthquake_distance.py
```diff
@@ -0,0 +1,179 @@
...
+                    "time": time,
+                    "distance_mi": dist_mi,
+                    "magnitude": mag,
+                }
+            )
+
+    return earthquake_distances
+
+
+# Example rune
```
tarnold17, line 173:
```suggestion
# Example
```

tarnold17, line 173:
```suggestion
# Example
```

pareto/utilities/earthquake_distance.py
```diff
@@ -0,0 +1,179 @@
...
+# swd_latlons = [
+#    {"id": 1, "lat": 34, "lon": -106},
+#    {"id": 2, "lat": 35, "lon": -105},
+# ]
+# earthquake_distances = calculate_earthquake_distances_texnet(swd_latlons)
```
tarnold17, line 178: I tried running this example with `calculate_earthquake_distances_usgs` instead of `calculate_earthquake_distances_texnet` and it returned an empty list. Can you provide an example of using `calculate_earthquake_distances_usgs` that does not return an empty list, so we can be sure that it is working as intended?

pareto/utilities/earthquake_distance.py
```diff
@@ -0,0 +1,179 @@
...
+import urllib.request
+import urllib.error
+import json
+import datetime
+import pyproj
+
+geod = pyproj.Geod(ellps="WGS84")
+mi_per_m = 0.000621371  # mi/m
+max_radius_mi = 5.59
+min_magnitude = 3
```
tarnold17, line 23: Instead of `max_radius_mi` and `min_magnitude` being fixed constants, can these be arguments that are passed to the function(s)? It makes sense to leave the default values as 5.59 and 3, respectively, but I think the user should be able to change these parameters if they wish.

tarnold17 (changes requested):
I like the work here! I have a few suggestions:

- The functions `calculate_earthquake_distances_usgs` and `calculate_earthquake_distances_texnet` are nearly identical, line for line. I suggest combining into one single function `calculate_earthquake_distances` and adding a keyword argument `database` where the user can pass either `"USGS"` or `"TexNet"` to select.
- Can you add an argument to the function(s) to allow the user to specify the date range for which they want to obtain results?
- Please write some unit tests for this utility. Note you will probably want to make use of pytest's `skipif` marker (https://docs.pytest.org/en/stable/how-to/skipping.html#id1) so that these tests only run for python version >= 3.9 (because pyproj is only supported for >= 3.9).
- Would it be possible to add an option to allow the user to export the results from the query in an Excel or csv format?
- I think it would be a good idea to a page to the [Utilities section](https://pareto.readthedocs.io/en/latest/utilities/index.html) of the PARETO documentation with documentation for this utility.

Let me know if you have questions about any of the above.

tarnold17:
Another request - please add documentation for this utility (perhaps in the Utilities/Get Data Functions section).

HuidaeCho:
> I like the work here! I have a few suggestions:
> 
>     * The functions `calculate_earthquake_distances_usgs` and `calculate_earthquake_distances_texnet` are nearly identical, line for line. I suggest combining into one single function `calculate_earthquake_distances` and adding a keyword argument `database` where the user can pass either `"USGS"` or `"TexNet"` to select.
> 

`api="usgs"` or `api="texnet"`

>     * Can you add an argument to the function(s) to allow the user to specify the date range for which they want to obtain results?
> 

`min_date` and `max_date` in `YYYY-MM-DD`; no constraints by default

>     * Please write some unit tests for this utility. Note you will probably want to make use of pytest's `skipif` marker (https://docs.pytest.org/en/stable/how-to/skipping.html#id1) so that these tests only run for python version >= 3.9 (because pyproj is only supported for >= 3.9).
>

Why do you use 3.9 as min version? My pyproj is 3.6.1 and works fine.
 
>     * Would it be possible to add an option to allow the user to export the results from the query in an Excel or csv format?
> 
>     * I think it would be a good idea to a page to the [Utilities section](https://pareto.readthedocs.io/en/latest/utilities/index.html) of the PARETO documentation with documentation for this utility.
> 
> 
> Let me know if you have questions about any of the above.

HuidaeCho:
> Another request - please add documentation for this utility (perhaps in the Utilities/Get Data Functions section).

Because this file is separate from `get_data.py`, I created a new document. https://pareto--305.org.readthedocs.build/en/305/utilities/Earthquake_Distance.html

tarnold17:
@HuidaeCho For testing, we use the pytest package. All of our existing unit tests are at `repos/project-pareto/tests`. Can you create a new file in this location called `test_earthquake_distance.py` with some tests? I think writing tests should be pretty simple - you can probably use the same examples that you have under `if __name__ == "__main__":` in `earthquake_distance.py`, and ensure that none of the calls to `calculate_earthquake_distances()` result in an error.

pareto/utilities/earthquake_distance.py
```diff
@@ -0,0 +1,230 @@
...
+                {
+                    keys[0]: swd_id,
+                    keys[1]: eq_id,
+                    keys[2]: time,
+                    keys[3]: dist_mi,
+                    keys[4]: mag,
+                }
+            )
+
+    if fmt == "csv":
```
tarnold17, line 187: I'm finding that the csv files generated have every other line blank, e.g.:

image

Is there a way to modify this so that we don't have the blank lines?

tarnold17 (commented):
Really nice work! I have just a couple more minor suggestions.

pareto/utilities/earthquake_distance.py
HuidaeCho: @tarnold17 Where do you see this problem? I'm using Windows newlines and it works for me.
image

HuidaeCho: The above screenshot is from VIM (^M is a Windows newline). This is from Notepad in Windows:
image

tarnold17: @HuidaeCho my results are from running `python earthquake_distance.py` and then opening the file that gets created `eq_dist_texnet_results.csv`. The screenshot I posted is from Notepad++. Here's what it looks like for me opening in Excel and Notepad:

image

image

So the issue seems apparent when viewing in Notepad++ or Excel, but not in Notepad. I am not entirely sure what to make of that.

tarnold17: @HuidaeCho here's what I see in Notepad++ when I turn on the "Show all characters" option: 
image

HuidaeCho: Got it. CR,CR,LF needs to be fixed to CR,LF.

pareto/tests/test_earthquake_distance.py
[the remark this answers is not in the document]
HuidaeCho: Will do after class.

tarnold17 (approved):
Great work!

tarnold17 (approved):

melody-shellman (approved):
Really nice, great new feature!

PR #319, opened 2024-04-25
Fix and reformat earthquake distance tests

tarnold17:

Reformat `pareto/tests/test_earthquake_distance.py` into multiple small tests instead of one large test, and mark with `@pytest.mark.xfail` the test which is failing in the GitHub environment.

Update codecov-action to v4 to hopefully resolve the persistent failures we're seeing using v2

### Legal Acknowledgement

contribution:

   employer.

### DIFF
--- a/pareto/tests/test_earthquake_distance.py
+++ b/pareto/tests/test_earthquake_distance.py
@@ -22,30 +22,38 @@ pytest.importorskip("pyproj")
 from pareto.utilities.earthquake_distance import calculate_earthquake_distances, main
 
 
-# calculate_earthquake_distances uses pyproj package, which is only supported
-# for Python 3.9 and later
-@pytest.mark.unit
-def test_earthquake_distance():
-    swd_latlons = [
-        {"swd_id": 1, "lat": 32.251, "lon": -101.940},
-        {"swd_id": 2, "lat": 31.651, "lon": -104.410},
-    ]
+swd_latlons = [
+    {"swd_id": 1, "lat": 32.251, "lon": -101.940},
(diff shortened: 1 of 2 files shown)


pareto/tests/test_earthquake_distance.py
```diff
@@ -22,30 +22,38 @@
...
+@pytest.mark.unit
+def test_earthquake_distance_usgs():
     # Ensure that request to USGS database works
     earthquake_distances = calculate_earthquake_distances(
         swd_latlons,
         api="usgs",
-        min_date="2024-03-23",
-        max_date="2024-03-23",
+        min_date="2023-11-14",
+        max_date="2024-04-14",  # 5 months
```
JavalVyas2000, line 48: The dates are 5 months apart for this particular test but in other tests it is the same data as 2024-03-23. Is there a reason to this change? Apart from this looks good to me.

JavalVyas2000 (approved):

ElmiraShamlou (dismissed):
LGTM

pareto/tests/test_earthquake_distance.py
tarnold17: @JavalVyas2000 it's essentially an arbitrary date range. We just want the range to be large enough that the query will reliably return at least one hit. Previously, the USGS database was returning 1 earthquake on 3/23/2024, but the underlying data in the database seems to have changed and that result is no longer returned. Using a larger range should make the test more robust going forward. We have not applied the same change to the test for the TexNet database, but we could change that in the future in a similar manner if necessary.

PR #324, opened 2024-05-17
Adding surrogate training notebook

JavalVyas2000:
This PR adds a notebook which demonstrates the workflow of training a keras machine learning surrogate.

### Legal Acknowledgement

contribution:

   employer.


pareto/examples/desalination_jupyter_notebooks/desalination_surrogate_training.ipynb
tarnold17, line 1: @JavalVyas2000 I get the following error when running the notebook:

`ModuleNotFoundError: No module named 'sklearn'`

Should we make sklearn a dependency of PARETO in this PR?

JavalVyas2000: Yes, that makes sense.

tarnold17: Seems this is no longer necessary - resolving.

tarnold17, line 1: I can't comment inline anymore because the file is too large, but I recommend removing these comments:

    # mvc_data.xlsx is a dummy file name.
    # One should replace this with the actual file name

Especially if we are able to add the data to this PR, because `mvc_data.xlsx` is in fact the name of the data file for MVC. You could add a different comment which says `# Replace "./mvc_data.xlsx" with "./md_data.xlsx" to train the model for MD instead of MVC`.

pareto/examples/desalination_jupyter_notebooks/desalination_surrogate_training.ipynb
```diff
@@ -0,0 +1,277 @@
...
+    "np.random.seed(46)\n",
+    "rn.seed(1342)\n",
+    "tf.random.set_seed(62)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "After importing the libraries, we will begin by reading the data generated from the Monte Carlo Simulation of the desalination model from an `csv` file. We will then remove any points where the model was found to be infeasible using the `dropna` function from pandas. Given that the simulations are on the order of 1e3, we will use the entire dataset; however, users can choose the number of samples they want to use for training the surrogate model. Finally, we will define the input and output variables for the surrogate model and extract the input and output labels from the column names of the dataset."
```
tarnold17, line 49:
```suggestion
    "After importing the libraries, we will begin by reading the data generated from the Monte Carlo Simulation of the desalination model from a `csv` file. We will then remove any points where the model was found to be infeasible using the `dropna` function from pandas. Given that the simulations are on the order of 1e3, we will use the entire dataset; however, users can choose the number of samples they want to use for training the surrogate model. Finally, we will define the input and output variables for the surrogate model and extract the input and output labels from the column names of the dataset."
```

tarnold17 (approved):
Looks good to me aside from one very minor additional comment!

tarnold17 (approved):

PR #327, opened 2024-06-11
Adding SCIP and GLPK functionality

JavalVyas2000:
## Motivation:

Adding SCIP and GLPK solver capabilities. There are still a few questions to be addressed. 

- [ ] SCIP requires an executable path. Where should we pass this executable path. 
- [ ] Check for other ways which may not require executable path for SCIP. 
- [ ] GLPK currently does not have time limit on it in the PR. 
- [ ] Add tests once they are ready to be merged. 

- Another note to take into account before we proceed with this PR is that the GLPK is significantly slower than any other solver. It did not go below 101% mipgap for the toy case study in a couple of minutes whereas SCIP solved it in 30 seconds. That is something to take into consideration, as the tests already pass an hour. This would make it even slower. 

### Legal Acknowledgement

contribution:

   employer.

### DIFF
--- a/pareto/strategic_water_management/strategic_produced_water_optimization.py
+++ b/pareto/strategic_water_management/strategic_produced_water_optimization.py
@@ -8188,7 +8188,9 @@ def solve_model(model, options=None):
         "gurobi_direct",
         "gurobi",
         "gams:CPLEX",
+        "scip",
         "cbc",
+        "glpk"
     )  # solvers to try and load in order
 
     # raise an exception if options is neither None nor a user-provided dictionary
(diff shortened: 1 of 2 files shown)


lbianchi-lbl:
@JavalVyas2000 to install SCIP in the CI environment, you can try adding `pip install --index-url https://pypi.ampl.com ampl-module-scip` just below L74 in the following file: https://github.com/project-pareto/project-pareto/blob/25852775369d9503764dd5570be87d8e507fe6e8/.github/workflows/checks.yml#L74

PR #336, opened 2024-07-02
Visualization of network. 

### LINKED ISSUE #321: Network Graph plot for visualizing feasibility of networks
> This is a suggestion, if possible plot the network graph using networkx. This will give a good visual representation of the model and make it possible for the user to check if there are any missing arcs without going through the excel sheet. I can help with this implementation. The following resource has a [tutorial](https://networkx.org/documentation/stable/tutorial.html) for the same.

_Originally posted by @JavalVyas2000 in https://github.com/project-pareto/project-pareto/pull/291#discussion_r1580615396_

JavalVyas2000:

Resolves #321 , the current version tries to mimic the schematics tab in the case study (which does not include the trucking arcs). 

- Adding visualize.py file. 
- Currently making a directional graph does not look pretty, and there are some overlaps, thus the graph represented is a bidirectional graph. 

The below is how the toy case study looks like. 

image

### Legal Acknowledgement

contribution:

   employer.


pareto/utilities/visualize.py
tarnold17, line 1: Why are only completions colored red and everything else is light green? I think that perhaps it makes sense to have a different color for every different network element.

tarnold17, line 1: Network elements in the graph figure are connected with a dotted line. Would it be possible to include an arrow for arcs which are unidirectional? That would be a nice touch. In other words, we would use "------" to connect bidirectional arcs and "----->" to connect unidirectional arcs.

tarnold17, line 1: Please add some tests for this feature.

pareto/utilities/visualize.py
```diff
@@ -0,0 +1,130 @@
...
+# itself and others acting on its behalf a paid-up, nonexclusive, irrevocable, worldwide license in
+# the Software to reproduce, distribute copies to the public, prepare derivative works, and perform
+# publicly and display publicly, and to permit others to do so.
+#####################################################################################################
+
+import networkx as nx
+import matplotlib.pyplot as plt
+
+
+def plot_network(m):
```
tarnold17, line 18: I'd recommend adding an option named "save" or possibly "filename" to this function to allow the user to save the figure that is generated. I think it makes sense for the default behavior to be that the file is not saved, but it would be nice to have this option available.

pareto/utilities/visualize.py
```diff
@@ -0,0 +1,130 @@
...
+    G = nx.Graph()
+    for i in m.s_L:
+        G.add_node(i)
+
+    piping_acrs = []
+    for i in m.s_LLA:
+        piping_acrs.append(i)
+    trucking_arcs = []
+    for i in m.s_LLT:
+        trucking_arcs.append(i)
```
tarnold17, line 28:
```suggestion
```
It seems that these lines of code can be deleted since trucking arcs are not plotted. However, you might be able to add an option to the function to allow the user to include trucking arcs in addition to or instead of piping arcs if desired. Does networkx allow you to draw two edges between nodes in a graph? Suppose two PW network elements are connected by a trucking arc and a piping arc. Could we then create a figure where there are two edges between these nodes, one for trucking and one for piping? We would need the edges to look different in some manner - e.g., we could use dashed lines for trucking arcs and solid lines for piping arcs.

pareto/utilities/visualize.py
tarnold17, line 1: Can you add a page to the "Utilities" section of the readthedocs with documentation for this feature?

tarnold17 (commented):
This is a really cool feature that I can see being very useful. I do have some questions and comments, however.

pareto/utilities/visualize.py
tarnold17, line 1: It seems that this feature is intended to be used to plot the entire network. But it would be nice to have an option to plot a graph which highlights the optimization results - i.e., a graph which omits infrastructure buildout options which are not executed.

pareto/examples/desalination_jupyter_notebooks/desalination_surrogate_training.ipynb
```diff
@@ -46,7 +46,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+
     "After importing the libraries, we will begin by reading the data generated from the Monte Carlo Simulation of the desalination model from a `csv` file. We will then remove any points where the model was found to be infeasible using the `dropna` function from pandas. Given that the simulations are on the order of 1e3, we will use the entire dataset; however, users can choose the number of samples they want to use for training the surrogate model. Finally, we will define the input and output variables for the surrogate model and extract the input and output labels from the column names of the dataset."
+
```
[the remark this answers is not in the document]
JavalVyas2000: I dont know how this changed. I shall revert back to the original version.

pareto/utilities/visualize.py
JavalVyas2000: I was trying to replicate the schematic in the excel, but could not accomplish the cross in the completions pad and hence decided to color it red. We can surely do different colors for different network elements.

pareto/utilities/visualize.py
JavalVyas2000: Thats a nice suggestion. I shall add that.

pareto/utilities/visualize.py
JavalVyas2000: I shall add that. Once the features are finalized we can add that.

JavalVyas2000: Yes, that can be done. I shall add another function that would visualize the results. I am not sure how would I visualize it using the same function.

JavalVyas2000: Yes, this is possible. The reason I did not have that is the plot starts to overlap. 

image

tarnold17: @JavalVyas2000 so simply using arrows instead of lines causes some of the elements of the graph to overlap? That is quite strange, and it might be worth investigating a fix, because I do think that using arrows to denote unidirectional arcs would be quite helpful.

tarnold17: @JavalVyas2000 I would think that a function for visualizing the results would probably have a good amount of overlap with the existing function, and that you might be able to reuse/repurpose some of the code that you have already written.

JavalVyas2000: I have added an argument to the function called show_results which is a boolean. If true, it shall show the results.

JavalVyas2000: I am not sure what tests would be feasible for this feature?

pareto/utilities/visualize.py
```diff
@@ -0,0 +1,208 @@
...
+    if show_results != None:
+        built_pipes = []
+        built_trucking = []
+        for t in m.s_T:
+            for i in piping_arcs:
+                if m.v_C_Piped[i, t].value:
+                    built_pipes.append(i)
+            for i in trucking_arcs:
+                if m.v_C_Trucked[i, t].value:
+                    built_trucking.append(i)
```
tarnold17, line 161: This code seems to not have any effect on the graph which is produced. I think perhaps we should punt for now on the idea of a "show_results" feature and possibly return to it in the future.

JavalVyas2000: Can we not just call draw_arcs function in this?

pareto/utilities/visualize.py
tarnold17: We now are using different shapes for different network elements, but everything is colored light green (except for freshwater sources which are light blue). We can revisit later to change more colors if we want, but for now I am resolving this thread.

tarnold17: Arrows are included now; resolving.

tarnold17: We have added tests which simply run the network visualization feature to make sure it doesn't throw any exceptions. Resolving this thread.

pareto/utilities/visualize.py
tarnold17: Resolved.

pareto/utilities/visualize.py
tarnold17: The option to plot trucking arcs is now implemented; resolving.

tarnold17 (approved):

Tominapa (approved):

PR #341, opened 2024-07-16
Don't print zero-valued variables in report

tarnold17:

#258 changed the `generate_report()` function in `get_data.py` to print zero-valued variables in the results report (previously, such variables were simply omitted from the report). It turns out that solved PARETO models tend to have LOTS of zero-valued variables - this change was causing the results objects for some models to be ~20x larger than before, which was degrading the performance of the UI. This PR reverts that change - it switches back to not printing zero-valued variables or expressions to results to keep the file size down.

### Legal Acknowledgement

contribution:

   employer.


MichaelPesce (approved):

ElmiraShamlou (approved):
LGTM

